PR #60, opened 2014-12-05
Refactor logging configuration

mouadino:
Move logging configuration to lymph.utils.logging and add a new function
setup_logger that can be used by specific drivers and lymph extentions
to setup logging for specific logger instead of having hardcoded logger
configuration in lymph.

### DIFF
--- a/lymph/utils/logging.py
+++ b/lymph/utils/logging.py
@@ -1,8 +1,8 @@
 from __future__ import absolute_import
-
 import logging
-import six
+from logging.config import dictConfig
 
+import six
 import zmq.green as zmq
 
 from lymph.utils.sockets import bind_zmq_socket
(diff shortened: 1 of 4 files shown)


lymph/utils/logging.py
```diff
@@ -25,12 +36,44 @@ def __init__(self, endpoint):
...
+        if isinstance(potentially_text, six.text_type):
+            return potentially_text.encode(encoding)
+        return potentially_text
 
 
-def _encode(potentially_text, encoding='utf-8'):
-    if isinstance(potentially_text, six.text_type):
-        return potentially_text.encode(encoding)
-    return potentially_text
+def setup_logging(logconf, loglevel, logfile, log_endpoint):
```
emulbreh: IMHO, this doesn't belong in utils (it's only useful for the cli).

emulbreh:
+1

PR #72, opened 2015-01-02
Fix variable name

mouadino:
After the refacoting in #70 Channel now accept a ZmqRPCServer instance and not the container anymore.


manisenkov:
:+1: :)

PR #73, opened 2015-01-02
Remove channel even when timeout

mouadino:
This should fix leaked memory where channel never get deleted from the
server.channels dictionary when Timeout.


manisenkov:
:+1:

PR #74, opened 2015-01-07
Add a flag that allow only showing running services

mouadino:


manisenkov:
Probably better to have this flag enabled by default.

mouadino:
I wanted to do that, but I do remember @emulbreh saying that this was conscious decision, so if you agree, let's do it this way and we can change it later if that is ok.

manisenkov:
Ah, ok. Then :+1: and think about this later

PR #79, opened 2015-01-13
Report all unhandled exception to sentry

mouadino:
This include all exception raised in greenlet and unhandled exception
that are going to crash the process.

### DIFF
--- a/lymph/core/container.py
+++ b/lymph/core/container.py
@@ -2,6 +2,7 @@ import json
 import gc
 import logging
 import os
+import sys
 
 import gevent
 import gevent.queue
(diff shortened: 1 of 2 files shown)


mamachanko:
:+1:

lymph/core/container.py
```diff
@@ -60,6 +61,9 @@ def __init__(self, ip='127.0.0.1', port=None, registry=None, logger=None, events
         if events:
             events.install(self)
 
+        # Set global exception hook to send unhandled exception to the container's error_hook.
+        sys.excepthook = self.excepthook
```
emulbreh: sys.excepthook should be set in the cli command so that we don't set it the tests.

mouadino: Good point, updated

PR #80, opened 2015-01-14
Fix logging tail

mouadino:
- Observer callback semantic changed.
- log_endpoint wasn't set correctly.
- move the excepthook after the logging setup to not endup with
  'No handlers could be found for logger "lymph"'

### DIFF
--- a/lymph/utils/logging.py
+++ b/lymph/utils/logging.py
@@ -23,15 +23,21 @@ def setup_logger(name):
     for hdlr in lymph_logger.handlers:
         logger.addHandler(hdlr)
     logger.setLevel(lymph_logger.level)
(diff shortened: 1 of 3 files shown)


lymph/utils/logging.py
```diff
@@ -27,11 +27,14 @@ def setup_logger(name):
 
 
 class PubLogHandler(logging.Handler):
-    def __init__(self, endpoint):
+    def __init__(self, endpoint, socket=None):
```
emulbreh: Let's mate ``endpoint`` optional and ``assert not (endpoint and socket)`` (or raise TypeError)

emulbreh: Maybe that's not as easy as I though with the dict config.

emulbreh:
+1

PR #84, opened 2015-01-16
Fix reporting errors when handling events

mouadino:
Change the event system to use container spawn and make sure that
any rejected message is reported to container error hook.


emulbreh:
+1

PR #85, opened 2015-01-16
Always return a Service instance in lookup

mouadino:
Don't make the return type depend on argument it's a bad practice
that yield many problems.


emulbreh:
+1

PR #86, opened 2015-01-16
Stop when there is nothing to tail

mouadino:


lymph/cli/tail.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import print_function
```
manisenkov: We can use six.print_ here, or just drop this import: print(...) will work even if we on python 2.7

mouadino: I have no idea why six has a print, but using __future__ print_function is IMHO the right way since it a real function e.g. you can do ``print('a', file=sys.stderr)`` :)

manisenkov:
:+1:

PR #89, opened 2015-01-26
Added basic newrelic plugin

emulbreh:


lymph/core/decorators.py
```diff
@@ -32,15 +32,12 @@ class RPCBase(collections.Callable):
     """
 
     def __init__(self, func, assigned=functools.WRAPPER_ASSIGNMENTS):
-        self._original = func
+        self.original = func
```
mouadino: what the reason of this change ? property are readonly now it's writtable, else it shouldn't ?

emulbreh: I wouldn't create properties for attributes just because they are supposed to be readonly – we'd basically have to write getters for most of our attributes then.

mouadino: I guess it's subjective and there is no need to argue on this, but is this change required for this PR ?

emulbreh: No, not necessary. Do I have to revert it?

mouadino: It will be great if we don't have to change others code just to fit our own code style that is considered subjective :) else we will end up with a big mess (talking from experience :)), thoughts ?

mouadino:
BTW we need to add newrelic to requirements right ?

emulbreh:
True, added it to extras_require.

lymph/core/decorators.py
emulbreh: I don't randomly change your code. If I'd do that, I would have renamed the decorator classes and removed the inheritance (I remember your original pull request). I just touched ``_original``, noticed that it was only used in one place and changed the property to an attribute because it's shorter and way cleaner.

PR #91, opened 2015-01-27
Add hostname to service info

MislavStipetic:


lymph/core/container.py
```diff
@@ -43,6 +44,7 @@ def __init__(self, ip='127.0.0.1', port=None, registry=None, logger=None, events
         self.node_endpoint = node_endpoint
         self.log_endpoint = log_endpoint
         self.service_name = service_name
+        self.hostname = platform.node()
```
mouadino: IMHO better to use fqdn: https://github.com/deliveryhero/lymph/blob/e08f481b5443da625b4983c17d3d0829b54b6b81/lymph/core/monitoring.py#L35

lymph/core/container.py
```diff
@@ -155,6 +157,7 @@ def get_instance_description(self, service_type=None):
             'endpoint': self.endpoint,
             'identity': self.identity,
             'log_endpoint': self.log_endpoint,
+            'hostname': self.hostname,
```
mouadino: Why didn't you change name hostname to fqdn ? since for some it can be confusing !

mouadino:
+1

PR #92, opened 2015-01-27
Enable coveralls

mouadino:


mouadino:
I am giving up on coveralls for now, it's seem that doesn't support re-test when re-pushing to a PR, and I dislike the fact that it play an intermediate between us and travis :(

PR #93, opened 2015-01-28
GUP-222 fixed heartbeat interval

mamachanko:
Rather than scheduling the next heartbeat dynamically we fix it to a value. Includes some refactoring. Removes Unicode character from logs. Heartbeat is virtually `ping`. Heartbeat is used to measure `rtt`.

### DIFF
--- a/lymph/core/connection.py
+++ b/lymph/core/connection.py
@@ -44,6 +44,8 @@ class Connection(object):
         self.heartbeat_loop_greenlet = self.server.container.spawn(self.heartbeat_loop)
         self.live_check_loop_greenlet = self.server.container.spawn(self.live_check_loop)
 
+        self.pid = os.getpid()
+
     def __str__(self):
         return "connection to=%s last_seen=%s" % (self.endpoint, self._dt())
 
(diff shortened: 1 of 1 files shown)


lymph/core/connection.py
```diff
@@ -74,35 +69,39 @@ def heartbeat_loop(self):
...
+                self.set_status(IDLE)
+                self.idle_since = now
+            else:
+                self.set_status(RESPONSIVE)
+
+    def log_stats(self):
+        roundtrip_stats = 'window (mean rtt={mean:.1f} ms; stddev rtt={stddev:.1f})'.format(**self.heartbeat_samples.stats)
+        roundtrip_total_stats = 'total (mean rtt={mean:.1f} ms; stddev rtt={stddev:.1f})'.format(**self.heartbeat_samples.total.stats)
+        logger.debug("pid=%s; %s; %s; phi=%.3f; ping/s=%.2f; status=%s" % (
+            os.getpid(),
```
mouadino: can we memoize this ? syscall are not cheap.

lymph/core/connection.py
```diff
@@ -74,35 +69,48 @@ def heartbeat_loop(self):
...
+            self.pid,
+            roundtrip_stats,
+            roundtrip_total_stats,
+            self.phi,
+            self.explicit_heartbeat_count / max(1, time.monotonic() - self.created_at),
+            self.status,
+        ))
+
+    @property
+    def pid(self):
```
mouadino: Can we just set it in ``__init__`` ?

mamachanko: silly me

mouadino:
LGTM

probably @emulbreh should take merge this one :)

lymph/core/connection.py
```diff
@@ -49,6 +44,8 @@ def __init__(self, server, endpoint, heartbeat_interval=1, timeout=1, idle_timeo
         self.heartbeat_loop_greenlet = self.server.container.spawn(self.heartbeat_loop)
         self.live_check_loop_greenlet = self.server.container.spawn(self.live_check_loop)
 
+        self.pid = os.getpid()
```
emulbreh: We can just remove the pid from the logs.

PR #94, opened 2015-01-29
Setting the reader to utf-8 for web handler; supporting py34

dushyant88:
When running a WebInterface on python3.4 a TypeError is raised when you try to access the `json` attribute of the interface.

```
TypeError: the JSON object must be str, not 'bytes'
```

We have to decode the stream to utf-8 to be able to support py34.

### DIFF
--- a/lymph/web/handlers.py
+++ b/lymph/web/handlers.py
@@ -1,3 +1,4 @@
+import codecs
 import json
 
 from werkzeug.exceptions import MethodNotAllowed
@@ -7,6 +8,9 @@ http_methods = ('get', 'post', 'head', 'options', 'put', 'delete')
 
 
 class RequestHandler(object):
+
+    stream_reader = codecs.getreader("utf-8")
+
     def __init__(self, interface, request):
         self.request = request
         self.interface = interface
@@ -19,7 +23,7 @@ class RequestHandler(object):
     def json(self):
         # FIXME: should we really keep a reference to the parsed body?
         if self._json is None:
-            self._json = json.load(self.request.stream)
+            self._json = json.load(self.stream_reader(self.request.stream))
         return self._json
 
     def dispatch(self, args):
--- a/lymph/web/handlers.py
+++ b/lymph/web/handlers.py
@@ -19,9 +19,10 @@ class RequestHandler(object):
         return [method.upper() for method in http_methods if callable(getattr(self, method))]
 
     def json(self):
-        # FIXME: should we really keep a reference to the parsed body?
-        request_is_json = "application/json" == self.request.mimetype
-        if request_is_json and self._json is None:
+        if not "application/json" == self.request.mimetype:
+            raise ValueError("The request Content-Type is not JSON")
+
+        if self._json is None:
             reader = codecs.getreader(self.request.charset)
             self._json = json.load(reader(self.request.stream))
         return self._json
--- a/lymph/web/handlers.py
+++ b/lymph/web/handlers.py
@@ -9,8 +9,6 @@ http_methods = ('get', 'post', 'head', 'options', 'put', 'delete')
 
 class RequestHandler(object):
 
-    stream_reader = codecs.getreader("utf-8")
-
     def __init__(self, interface, request):
         self.request = request
         self.interface = interface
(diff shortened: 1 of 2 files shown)


lymph/web/handlers.py
```diff
@@ -19,7 +23,7 @@ def allowed_methods(self):
     def json(self):
         # FIXME: should we really keep a reference to the parsed body?
         if self._json is None:
-            self._json = json.load(self.request.stream)
+            self._json = json.load(self.stream_reader(self.request.stream))
```
mouadino: Not sure why we use stream in the first place since it's low level, IMHO better use ``json.load(self.request.data)`` which does already the right thing https://github.com/mitsuhiko/werkzeug/blob/master/werkzeug/wrappers.py#L444

mouadino: And we should probably follow the docstring advice, by checking first the content-type and only load if it's json.

emulbreh: I think working with streams would be cleaner. But I agree that we should take the charset from the request. If we use ``get_data()`` we must also set ``encoding_errors = 'strict'`` on the request.

mouadino: I think Werkzeug default may be better since you don't want to blow up with 500 in case user send wrong data, but leave the chance to app level to return a 400 with a more meaningful message.

emulbreh: yes, that's why I think that 'strict' would be better. Then you can catch the exception and return a 400 response instead of hiding the original problem (badly encoded request) with a json parser error.

mouadino: Then if you are going to do it here, the 400 response will not match developers use case, e.g. developer want to return 400 content-type is JSON and containing some custom message ... . 

Which mean that we need to give developer a way to supply their own custom exception that they want to raise and then figure out what should we pass to it ctor ... .

I still think Werkzeug choice of not using strict encoding error is right and that we should just use  Werkzeug API instead of reinventing ours, which is also the easiest thing we can also do and I love easy (Keep It Simple and Stupid) :)

But no-blocking, if you still thing that the right choice go ahead, but we should definitely add the check for content type before trying to load it.

emulbreh: http://werkzeug.pocoo.org/docs/0.9/unicode/#error-handling

lymph/web/interfaces.py
```diff
@@ -15,6 +16,10 @@
 logger = logging.getLogger(__name__)
 
 
+class Request(DynamicCharsetRequestMixin, BaseRequest):
```
dushyant88: Should I move this Request Class to another file - ```wrappers```?

lymph/web/handlers.py
```diff
@@ -19,7 +21,8 @@ def allowed_methods(self):
     def json(self):
         # FIXME: should we really keep a reference to the parsed body?
         if self._json is None:
-            self._json = json.load(self.request.stream)
+            reader = codecs.getreader(self.request.charset)
```
mouadino: Can you add a check for content-type ?

dushyant88: What kind of check? like if it is text?

mouadino: Like only load json if the content-type is json http://werkzeug.pocoo.org/docs/0.9/request_data/#how-to-extend-parsing

dushyant88: Aah;

But I think this will break our API + Tests - both of them don't set content-type at all!

lymph/web/handlers.py
```diff
@@ -18,8 +20,10 @@ def allowed_methods(self):
 
     def json(self):
         # FIXME: should we really keep a reference to the parsed body?
```
mouadino: Is this FIXME still valid ? @emulbreh ?

lymph/web/handlers.py
```diff
@@ -18,8 +20,10 @@ def allowed_methods(self):
...
-        if self._json is None:
-            self._json = json.load(self.request.stream)
+        request_is_json = "application/json" == self.request.mimetype
+        if request_is_json and self._json is None:
+            reader = codecs.getreader(self.request.charset)
+            self._json = json.load(reader(self.request.stream))
```
emulbreh: Let's raise a ValueError if the request is not json.

lymph/web/interfaces.py
emulbreh: It should probably be somewhere else, but I don't know where. Good enough for me.

lymph/web/handlers.py
emulbreh: can be removed, I made up my mind.

lymph/web/handlers.py
dushyant88: Wouldn't `json.load` raise a ValueError if it can't load json?

Or do you mean, we should raise a ValueError if the content-type is not application/json?

emulbreh: In the wrong-content-type-case. Currently it would just return None.

PR #95, opened 2015-02-04
Fix first time used connection

mouadino:
A newly created connection will have the UNKONWN state so until the heartbeat
greenlet get a chance to run, we will assume that this connection is alive.

This will fix first requests to services instead of timeouting.

### DIFF
--- a/lymph/core/connection.py
+++ b/lymph/core/connection.py
@@ -2,13 +2,14 @@
 from __future__ import division, unicode_literals
 
 import gevent
+from gevent.event import Event
 import math
 import os
 import time
 import logging
 
 from lymph.utils import SampleWindow
-from lymph.exceptions import RpcError
+from lymph.exceptions import Timeout
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


lymph/core/connection.py
```diff
@@ -125,7 +125,7 @@ def on_send(self, msg):
         self.sent_message_count += 1
 
     def is_alive(self):
-        return self.status in (RESPONSIVE, IDLE)
+        return self.status in (RESPONSIVE, IDLE, UNKNOWN)
```
mamachanko: How about only allowing `UNKOWN` to be considered alive if there was no heartbeat yet, i.e. `self.explicit_heartbeat_count` is 0?

mouadino: it's the current state, UNKOWN is only set as initial value for the connection so as soon as heartbeat start the state change to something else.

mamachanko:
``` python
def is_alive(self):
    if self.explicit_heartbeat_count == 0:
        return self.status in (RESPONSIVE, IDLE, UNKNOWN)
    return self.status in (RESPONSIVE, IDLE)
```

emulbreh:
+1, good enough for now

mouadino:
I give up ! too brittle code :(

PR #96, opened 2015-02-05
Fix discover method interface

mouadino:
The discover method is not supposed to take an argument as it is also
specified in this interface implementation.


emulbreh:
+1

PR #101, opened 2015-02-12
Move imported objects to separate function.

vpikulik:
This change will allow to inherit this command and write shell for other service with needed variables.

### DIFF
--- a/lymph/cli/shell.py
+++ b/lymph/cli/shell.py
@@ -20,10 +20,12 @@ class ShellCommand(Command):
 
     short_description = 'Open an interactive Python shell.'
 
-    def run(self, **kwargs):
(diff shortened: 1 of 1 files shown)


lymph/cli/shell.py
```diff
@@ -20,10 +20,12 @@ class ShellCommand(Command):
 
     short_description = 'Open an interactive Python shell.'
 
-    def run(self, **kwargs):
+    def _get_imported_objects(self, **kwargs):
```
jacqueslh: if you plan to reuse if in other classes, the `_` isn't needed right ?

vpikulik: you right. changing

jacqueslh:
:+1:

PR #104, opened 2015-02-17
Add the container service name to the process name

kpanic:
A service should export his service_name to be more easily traced via `ps`


jacqueslh:
LGTM

PR #110, opened 2015-02-23
Add a better way to guess external ip

mouadino:
This doesn't need a DNS resolver to be set up or an external
request to be made which fix the problem in case a node is
not connected to internet.

### DIFF
--- a/lymph/cli/main.py
+++ b/lymph/cli/main.py
@@ -27,9 +27,10 @@ def setup_config(args):
     ip = args.get('--ip')
     if args.get('--guess-external-ip'):
         if ip:
-            print('cannot combine --ip and --guess-external-ip')
-            return 1
+            sys.exit('Cannot combine --ip and --guess-external-ip')
         ip = guess_external_ip()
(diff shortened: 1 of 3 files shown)


jacqueslh:
LGTM

lymph/cli/main.py
```diff
@@ -27,9 +27,10 @@ def setup_config(args):
     ip = args.get('--ip')
     if args.get('--guess-external-ip'):
         if ip:
-            print('cannot combine --ip and --guess-external-ip')
-            return 1
+            sys.exit('Cannot combine --ip and --guess-external-ip')
         ip = guess_external_ip()
+        if ip is None:
+            sys.exit('Cannot guess external ip, aboting ...')
```
eguven: `aboting` :)

eguven:
btw, this returns my gateway on laptop and not my IP on network. checked a few different interface configurations, looks good for them.

mouadino:
@eguven  thanks, fixed

PR #112, opened 2015-02-23
Dump threads stack when SIGQUIT is send to process

mouadino:
SIGQUIT signal is meant to be used to produce a core dump when it terminates
the process, JAVA virtual machines already support this behaviour, this change
add it to lymph. This is very usefull feature when it come to test why a process
hang, and specially in a production environment, where when this happen ops can
just kill it using ``kill -3 <pid>`` which will produce a snapshot of all
greenlets and os-threads stack printed stderr.

### DIFF
--- a/lymph/cli/service.py
+++ b/lymph/cli/service.py
@@ -1,12 +1,14 @@
-import gevent
 import logging
 import os
+from functools import partial
 import signal
-import six
 import sys
-from setproctitle import setproctitle
 
-from lymph.utils import import_object
+import gevent
+from setproctitle import setproctitle
+import six
+
+from lymph.utils import import_object, dump_stacks
 from lymph.autoreload import set_source_change_callback
 from lymph.cli.base import Command
 from lymph.core.container import create_container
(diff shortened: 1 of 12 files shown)


lymph/cli/service.py
```diff
@@ -50,45 +52,59 @@ class InstanceCommand(Command):
...
+        install_plugins(self.container, self.config.get('plugins', {}))
+        install_interfaces(self.container, self.config.get('interfaces', {}))
+
+        for cls_name in self.args.get('--interface', ()):
+            cls = import_object(cls_name)
+            self.container.install(cls)
+
+    def _start_backdoor_terminal(self):
+        from gevent.backdoor import BackdoorServer
+        backdoor = BackdoorServer(('127.0.0.1', 5005), locals={'container': self.container, 'dump_stacks': dump_stacks})
```
kashopi: Shouldn't this 5005 be in the config?

MislavStipetic: Could we add config via command line, so when we start the service in .lymph.yml, we can have multiple sessions instead of only one getting the port?

But that's non-blocking, +1

eguven: :+1: for configurable

lymph/services/node.py
```diff
@@ -29,9 +29,13 @@ def start(self):
             self.cmd, env=self.env, close_fds=False)
         self._process = psutil.Process(self._popen.pid)
 
-    def stop(self):
+    def stop(self, **kwargs):
```
emulbreh: ``def stop(self, signalnum=None, **kwargs)``

emulbreh:
+1

lymph/cli/service.py
mouadino: That my next task so not in this PR :)

PR #114, opened 2015-02-25
Allow shared sockets config to be a sequence instead of a mapping

emulbreh:


lymph/config.py
```diff
@@ -84,14 +87,21 @@ def get_instance(self, key, default_class=None, **kwargs):
...
-        return values[path[-1]]
+                if default is not Undefined:
+                    return default
+                raise KeyError()
+        try:
+            return values[path[-1]]
+        except KeyError:
+            if default is not Undefined:
+                return default
+            raise KeyError()
```
mouadino: maybe you want empty ``raise`` to reraise the above KeyError so that we can also see the key

lymph/config.py
```diff
@@ -84,14 +87,21 @@ def get_instance(self, key, default_class=None, **kwargs):
...
         path = key.split('.')
         values = self.values
         for bit in path[:-1]:
             values = values[bit]
             if values is None:
-                raise KeyError
-        return values[path[-1]]
+                if default is not Undefined:
+                    return default
+                raise KeyError()
```
mouadino: ``raise KeyError(bit)`` ?

lymph/config.py
emulbreh: I don't want to reraise, but we could add the full key to the error message.

mouadino:
LGTM

mamachanko:
example config?

PR #115, opened 2015-02-25
Provide context to lymph proxy

kpanic:
Use case: provide some context data to another service

### DIFF
--- a/lymph/core/interfaces.py
+++ b/lymph/core/interfaces.py
@@ -37,6 +37,7 @@ class InterfaceBase(type):
         new_cls.declarations = declarations
         return new_cls
 
+
 class ProxyMethod(object):
     def __init__(self, func):
         self.func = func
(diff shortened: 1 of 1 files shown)


mouadino:
How about if remote function have an argument named 'context' ?

lymph/core/interfaces.py
```diff
@@ -51,16 +52,20 @@ def defer(self, *args, **kwargs):
...
         self._address = address
         self._method_cache = {}
         self._timeout = timeout
         self._namespace = namespace or address
         self._error_map = error_map or {}
+        self._context = context or {}
 
     def _call(self, __name, **kwargs):
-        channel = self._container.send_request(self._address, __name, kwargs)
+        kwargs['context'] = self._context
```
mouadino: How about if remote function have an argument named 'context' ?

emulbreh: yes, this should be explicit.

PR #116, opened 2015-02-25
Add ``--remote`` option to ``lymph shell``

mouadino:
1. Fix the backdoor terminal setup by removing the hardcoded port so
that we can have more than one service (and it instances) running
with a backdoor terminal, and make the backdoor listen ip configurable.

2. Make backdoor terminal endpoint discoverable by saving it to
zookeeper.

3. Add ``lymph terminal`` that open a telnet connection to remote
backdoor terminal of a give service.

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -93,10 +93,31 @@ This command takes the same commandline options as ``lymph instance``.
 ``lymph shell``
 ~~~~~~~~~~~~~~~
 
-Starts an interactive Python shell. The following objects will be available in the global namespace:
+Starts an interactive Python shell, locally or remotely.
+
+Locally:
+--------
+
+In case shell was open locally the following objects will be available in the
+global namespace:
 
 ``client``
     a configured :class:`lymph.client.Client` instance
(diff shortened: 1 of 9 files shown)


docs/cli.rst
```diff
@@ -100,3 +100,9 @@ Starts an interactive Python shell. The following objects will be available in t
 
 ``config``
     a loaded :class:`lymph.config.Configuration` instance
+
+
+``lymph terminal``
```
MislavStipetic: I don't want to be pedantic with names again, but lymph terminal sounds a bit like lymph shell. I would maybe call it lymph backdoor (since it called like that everywhere in the code). Non blocking of course.

mouadino: Not sure about the name, let's just wait to see what others make of this.

manisenkov: `backdoor` or `remote` are better for me. I'm agree with @MislavStipetic that `terminal` is confused with `shell`.

MislavStipetic:
+1

PR #117, opened 2015-02-25
Fix log format of web to be more splunky

mouadino:
Add also a unit for duration for us poor human that cant' compile
seconds milliseconds faster :)

### DIFF
--- a/lymph/core/rpc.py
+++ b/lymph/core/rpc.py
@@ -180,7 +180,7 @@ class ZmqRPCServer(object):
                 logger.exception('failed to send automatic NACK')
         finally:
             elapsed = (time.time() - start) * (10 ** 3)
-            logger.log(loglevel, '%s duration=%.3f', msg.subject, elapsed)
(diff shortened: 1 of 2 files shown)


lymph/core/rpc.py
```diff
@@ -180,7 +180,7 @@ def dispatch_request(self, msg):
                 logger.exception('failed to send automatic NACK')
         finally:
             elapsed = (time.time() - start) * (10 ** 3)
-            logger.log(loglevel, '%s duration=%.3f', msg.subject, elapsed)
+            logger.log(loglevel, '%s duration=%.3f (seconds)', msg.subject, elapsed)
```
vpikulik: 'subject=%s duration=%.3f (seconds)' it should be better for splunk

mamachanko: quote it if in doubt

mouadino: quote what ?

vpikulik:
:+1:

PR #118, opened 2015-03-02
Add support for one-time events

manisenkov:
Sample of usage:

```python
class ConsumerService(lymph.Interface):
    @lymph.rpc()
    def consume(self, **kwargs):
        ident = str(uuid4())
        async_result = self.get_next_event('onetime_event.%s' % ident)

        # Doing something that cause event
        self.proxy('producer').produce(ident=ident)

        result = async_result.get()  # Wait until event will be processed

        return {
            'source_id': ident,
            'target_id': result['target_id'],
            'generated_id': result['generated_id']
        }


class ProducerService(lymph.Interface):
    @lymph.rpc()
    def produce(self, ident, **kwargs):
        self.emit('onetime_event.%s' % ident, {
            'target_id': ident,
            'generated_id': str(uuid4())
        })
```

### DIFF
--- a/lymph/core/events.py
+++ b/lymph/core/events.py
@@ -1,5 +1,6 @@
 import re
 import logging
+from uuid import uuid4
 
 from lymph.core.interfaces import Component
 from lymph.core import trace
@@ -42,17 +43,22 @@ class Event(object):
 
 
 class EventHandler(Component):
-    def __init__(self, interface, func, event_types, sequential=False, queue_name=None, active=True):
+    def __init__(self, interface, func, event_types, sequential=False, queue_name=None, active=True, once=False):
         self.func = func
         self.event_types = event_types
         self.sequential = sequential
         self.active = active
         self.interface = interface
+        self.once = once
+        self.unique_key = str(uuid4()) if once else None
         self._queue_name = queue_name or func.__name__
 
     @property
     def queue_name(self):
-        return '%s-%s' % (self.interface.name, self._queue_name)
+        if self.unique_key:
+            return '%s-%s-%s' % (self.interface.name, self._queue_name, self.unique_key)
+        else:
+            return '%s-%s' % (self.interface.name, self._queue_name)
 
     @queue_name.setter
     def queue_name(self, value):
--- a/lymph/core/interfaces.py
+++ b/lymph/core/interfaces.py
@@ -3,11 +3,25 @@ import functools
 import six
 
 from lymph.core.decorators import rpc, RPCBase
-from lymph.exceptions import RemoteError
+from lymph.exceptions import RemoteError, EventHandlerTimeout
 from lymph.core.declarations import Declaration
 
 import gevent
 from gevent.event import AsyncResult
+
+
+class AsyncResultWrapper(object):
+    def __init__(self, container, handler, async_result):
+        self.container = container
+        self.handler = handler
+        self.result = async_result
+
+    def get(self, timeout=30):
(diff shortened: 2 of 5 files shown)


lymph/events/kombu.py
```diff
@@ -48,6 +48,9 @@ def message_handler():
                 # Since the message handler can be run sequentially, we are catching all exception
                 # and reporting them here.
                 self.event_system.container.error_hook(sys.exc_info())
+            finally:
+                if self.handler.once:
```
mouadino: We are rejecting message when we can't processed them, that may come as a surprise, maybe we want to requeue them in this case we should not delete queue in finally !

lymph/core/interfaces.py
```diff
@@ -3,13 +3,27 @@
...
 from gevent.event import AsyncResult
 
 
+class UnsubscribingResult(object):
+    def __init__(self, container, handler, result):
+        self.container = container
+        self.handler = handler
+        self.result = result
+
+    def get(self, timeout):
```
mouadino: should we have a default timeout ?

jacqueslh: +1

mamachanko:
I admit I don't get it.

mamachanko:
Maybe we could provide some API to help the client wrap these three lines into something more concise:
``` python
    async_result = self.subscribe_async('onetime_event.%s' % ident)

    # Doing something that cause event
    self.proxy('producer').produce(ident=ident)

    result = async_result.get()  # Wait until event will be processed
```
instead:
``` python
    
```

(non-blocking)

lymph/core/interfaces.py
```diff
@@ -119,11 +134,24 @@ def proxy(self, address, **kwargs):
     def subscribe(self, *event_types, **kwargs):
         def decorator(func):
             from lymph.core.events import EventHandler
-            handler = EventHandler(func, event_types, **kwargs)
+            handler = EventHandler(self, func, event_types, **kwargs)
             self.container.subscribe(handler)
             return handler
         return decorator
 
+    def subscribe_once(self, *event_types, **kwargs):
```
emulbreh: Do we really need this? We can use `subscribe(..., once=True)` instead.

lymph/core/interfaces.py
```diff
@@ -119,11 +134,24 @@ def proxy(self, address, **kwargs):
...
+        kwargs['once'] = True
+        return self.subscribe(*event_types, **kwargs)
+
+    def subscribe_async(self, *event_types, **kwargs):
```
emulbreh: `get_next_event()`, not particularly nice, but at least emphasizes that it can only gets the next event.

lymph/core/interfaces.py
MislavStipetic: +1 for once=True argument

manisenkov: ok, agree with that

lymph/core/events.py
```diff
@@ -42,17 +43,22 @@ def serialize(self):
...
         self.active = active
         self.interface = interface
+        self.once = once
+        self.unique_key = str(uuid4()) if once else None
         self._queue_name = queue_name or func.__name__
 
     @property
     def queue_name(self):
-        return '%s-%s' % (self.interface.name, self._queue_name)
+        if self.once:
```
emulbreh: `if self.unique_key`

emulbreh:
+1

PR #122, opened 2015-03-09
compute elapsed time in seconds for RPC call

mamachanko:
* `time.time()` is returning seconds already (https://docs.python.org/2/library/time.html#time.time)


mouadino:
LGTM

PR #127, opened 2015-03-16
Add a way to configure dependencies from configuration

mouadino:


docs/configuration.rst
```diff
@@ -3,16 +3,40 @@ Configuration
...
         hostname: 127.0.0.1
 
-You can find this sample configuration file in :file:`conf/service.yml`.
+You can find this sample configuration file in :file:`conf/sample-node.yml`.
+
+
+Dependencies
+------------
+
+Lymph support a way to inject dependencies from configuration file.
```
jacqueslh: `Lymph supports`

docs/configuration.rst
```diff
@@ -3,16 +3,40 @@ Configuration
...
+
+You start by defining a top level dependencies that you want to inject
```
jacqueslh: `a top level dependency` or `top level dependencies` ?

lymph/cli/testing.py
```diff
@@ -65,14 +67,11 @@ class CliTestMixin(object):
...
-        if not hasattr(self, '_cli'):
-            self._cli = CliWrapper(self.cli_config)
-        return self._cli
+        return CliWrapper(self.cli_config)
 
-    def tearDown(self):
-        if hasattr(self, '_cli'):
-            self._cli.tear_down()
-        super(CliTestMixin, self).tearDown()
+    # FIXME (Mouad): def tearDown(self):
```
emulbreh: what happened here?

docs/configuration.rst
mouadino: it should have been a top level "dependencies" key.

lymph/cli/testing.py
mouadino: A storm happen :)

We can't have one cli client per test case (this one took me a while to found :)) so that  explain the property change, the cli tear_down basically remove temp files which is good to have but not necessary so I prefer if I do it later on.

docs/configuration.rst
mouadino: Fixed

lymph/config.py
```diff
@@ -1,12 +1,16 @@
...
 
 import abc
+import copy
 import six
 import yaml
 
 from lymph.utils import import_object, Undefined
 
 
+_INSTANCES_CACHE = {}
```
emulbreh: that shouldn't be on the module level, can't we keep it on the Configuration object?

lymph/config.py
```diff
@@ -32,10 +36,19 @@ def setdefault(self, key, default):
...
+        instance_config = self.get(key, {})
+        return self._create_instance(instance_config, default_class=default_class, **kwargs)
+
+    def _create_instance(self, instance_config, default_class=None, **kwargs):
+        clspath = instance_config.get('class', default_class)
+        cls = import_object(clspath)
+        if hasattr(cls, 'from_config'):
+            return cls.from_config(instance_config, **kwargs)
+        else:
+            instance_config = copy.deepcopy(instance_config)
```
emulbreh: That's a deepcopy of a ConfigView. It seems to work (since it also copies the referenced Configuration instance), but I'd rather just copy the values: `deepcopy(dict(instance_config))`

lymph/discovery/zookeeper.py
```diff
@@ -15,28 +15,21 @@
...
         self.start_count = 0
 
     @classmethod
     def from_config(cls, config, **kwargs):
-        return cls(
-            hosts=config.get('hosts', DEFAULT_HOSTS),
-            chroot=config.get('chroot', DEFAULT_CHROOT),
-            **kwargs
-        )
+        zkclient = config.root.get_instance('registry.zkclient', handler=SequentialGeventHandler())
```
emulbreh: can't we use `config.get_instance('zkclient')` here? That wouldn't require knowledge of the config structure.

lymph/patterns/serial_events.py
```diff
@@ -13,22 +14,26 @@
...
-def serial_event(*event_types, **kwargs):
+def serial_event(key_func, *event_types, **kwargs):
+    if not event_types:
+        raise ValueError('A list of event types is required')
+
     def decorator(func):
         def factory(interface):
-            return SerialEventHandler(interface, func, event_types, **kwargs)
+            zkclient = interface.config.root.get_instance(
+                'components.SerialEventHandler.zkclient',
```
emulbreh: Not sure about this, but doesn't matter for now. The component config idea was to have a config section per `SerialEventHandler` instance.

lymph/testing/__init__.py
```diff
@@ -142,12 +143,18 @@ class LymphIntegrationTestCase(KazooTestHarness):
...
         if self.use_zookeeper:
             self.setup_zookeeper(handler=SequentialGeventHandler())
 
     def tearDown(self):
+        super(LymphIntegrationTestCase, self).tearDown()
         if self.use_zookeeper:
             self.teardown_zookeeper()
+        for container in self._containers:
+            container.stop()
+        _INSTANCES_CACHE.clear()
```
emulbreh: This would be unnecessary with a Configuration local instance cache.

lymph/patterns/serial_events.py
```diff
@@ -13,22 +14,26 @@
 logger = logging.getLogger(__name__)
 
 
-def serial_event(*event_types, **kwargs):
+def serial_event(key_func, *event_types, **kwargs):
```
emulbreh: Let's keep the signature here. It's already documented.

lymph/config.py
mouadino: Than it's an instance cache, which mean if I create to Configuration instance calling get_instance or get_dependency will not get the cached value ! is that what you want ?

lymph/config.py
mouadino: ConfigView is a Mapping, isn't it ? that's why deepcopy work, AFAIK dict(mapping) is just redundant is like doing dict(dict({...})) which is basically create a shallow copy, so deepcopy(dict(mapping)) will create a shallow copy than a deep copy for no good reason !

lymph/discovery/zookeeper.py
mouadino: Sadly no, I think this need more refactoring if we want it to work this way :(

lymph/patterns/serial_events.py
mouadino: not sure I follow, but probably this is related to the previous comment.

lymph/testing/__init__.py
mouadino: See comment about why local instance cache my not be a good idea.

lymph/patterns/serial_events.py
mouadino: it's a broken signature, documentation doesn't fix code !

emulbreh:
+1

PR #128, opened 2015-03-17
Fix lazy declaration setup

mouadino:
Some declaration are meant to be instantiated after apply_config others
are accessed immediately, this allow us to lazily evaluate declaration
and only when they are accessed.


jacqueslh:
LGTM after travis

PR #130, opened 2015-03-17
Fix configuration passed to constructor and from_config

mouadino:
We should pass ConfigView instead of dictionary in from_config since
some services require it.


mamachanko:
+1

PR #131, opened 2015-03-18
Add interface parameter for serial_event key function

manisenkov:


mamachanko:
cool +1

emulbreh:
This should be reflected in the docs as well, but +1

PR #132, opened 2015-03-18
Add enability to bound greenlets pool from configuration

mouadino:
Add a NonBlockingPool that in contrast of gevent.pool.Pool it's non blocking and
as soon as pool limit are reached it will reject work instead of blocking, which
can be very dangerous when you can block for ever and with it block calling greeenlet
in other services ... .

When Pool reject jobs it raise a ``ResourceExhausted`` subclass, the idea is to expose
this exception and then add a back pressure support with bulkhead pattern into client
for more resilience.

The NonBlockingPool is also used in gevent WSGI server to bound resources.

### DIFF
--- a/lymph/core/container.py
+++ b/lymph/core/container.py
@@ -39,7 +39,7 @@ class ServiceContainer(object):
 
     server_cls = ZmqRPCServer
 
-    def __init__(self, ip='127.0.0.1', port=None, registry=None, logger=None, events=None, node_endpoint=None, log_endpoint=None, service_name=None, debug=False, monitor_endpoint=None):
+    def __init__(self, ip='127.0.0.1', port=None, registry=None, logger=None, events=None, node_endpoint=None, log_endpoint=None, service_name=None, debug=False, monitor_endpoint=None, pool_size=None):
         self.server = self.server_cls(self, ip, port)
         self.node_endpoint = node_endpoint
         self.log_endpoint = log_endpoint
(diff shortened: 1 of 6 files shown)


lymph/core/container.py
```diff
@@ -39,7 +39,7 @@ class ServiceContainer(object):
 
     server_cls = ZmqRPCServer
 
-    def __init__(self, ip='127.0.0.1', port=None, registry=None, logger=None, events=None, node_endpoint=None, log_endpoint=None, service_name=None, debug=False, monitor_endpoint=None):
+    def __init__(self, ip='127.0.0.1', port=None, registry=None, logger=None, events=None, node_endpoint=None, log_endpoint=None, service_name=None, debug=False, monitor_endpoint=None, pool_size=None):
```
mamachanko: so many kwargs.

mouadino: yep and I hate it :)

mouadino:
As a next step, the idea is to split pools to:

- rpc_pool: RPC requests greenlets pool, should be by interface.
- wsgi_pool: Included in this patch.
- core_pool: Pool for other stuff like connections keepalive and so on ... this doesn't have to be configurable.

mamachanko:
lgtm

emulbreh:
I still don't quite understand the point of these limits. Do you have a benchmark that shows an improvement?

emulbreh:
+1

PR #135, opened 2015-03-22
Fixed discovery after zk disconnect

emulbreh:


mouadino:
It will be great if you can squash the commits since they are broken, but other than that LGTM

PR #136, opened 2015-03-23
pin the coverage version

stefano-dh:


requirements/dev.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/nedbat/coveragepy.git
+https://github.com/nedbat/coveragepy/archive/coverage-4.0a5.tar.gz#egg=coverage==4.0a5
```
stefano-dh: yes, I did not notice! this is even better!

PR #137, opened 2015-03-23
Make RpcMock test case a standalone class

mouadino:
This will make having rpc calls mocked optional for developers to
either include in there test cases or no.


emulbreh:
+1

PR #138, opened 2015-03-23
remove redundant requirements

stefano-dh:


emulbreh:
We should keep the other line, base requirements are needed.

stefano-dh:
@emulbreh those can be installed with pip install -e . as well though, right? I see there is some code in setup.py

emulbreh:
setup.py isn't used for pinning, that's why we have requirements/base.txt

PR #139, opened 2015-03-23
Allow specifying import path as a rule endpoint

mouadino:
Import path must be relativ to where Interface module is, example if current
web interface is in 'pkg/interfaces.py' then Rule(..., endpoint='handlers:SomeHandler')
will try to import handler in 'pkg.handlers:SomeHandler'.

### DIFF
--- a/docs/api/web_api.rst
+++ b/docs/api/web_api.rst
@@ -10,15 +10,48 @@ Web API
 
         If there's no port provided in the interface config, the http interface
         is bound to this port.
-    
+
     .. attribute:: application
-    
+
         WSGI application instance that this interface is running
-        
+
     .. attribute:: url_map
-    
+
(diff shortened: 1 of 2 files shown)


mamachanko:
What about tests and documentation?

mouadino:
@mamachanko Sadly web interface is lacking both, it's like a hidden feature :)

Writing tests for web interface from scratch part may take more time so I would like to postpone it, as for documentation let me see what I can cook.

mamachanko:
@mouadino I've created stories for both http://chapterbe1.lab:9021/p/PlatformPlanning#lineNumber=258

If you'd like to go ahead for preliminary documentation that'd be great though, I think.

mamachanko:
lgtm

mouadino:
Going with https://github.com/deliveryhero/lymph/pull/141/files instead

PR #140, opened 2015-03-23
Component refactoring

emulbreh:


lymph/core/components.py
```diff
@@ -0,0 +1,72 @@
...
+        if factory in self._declared_components:
+            raise RuntimeError("already installed: %s" % factory)
+        component = factory(self, **kwargs)
+        self._declared_components[factory] = component
+        self.add_component(component)
+        return component
+
+    def on_start(self):
+        for declaration in self.declarations:
+            declaration.__get__(self, type(self))
```
mamachanko: Isn't there a more "natural" way to do this?

lymph/core/components.py
```diff
@@ -0,0 +1,72 @@
...
+            if isinstance(value, Declaration):
+                value.name = name
+                declarations.add(value)
+        new_cls = super(ComponentizedBase, cls).__new__(cls, clsname, bases, attrs)
+        new_cls.declarations = declarations
+        return new_cls
+
+
+@six.add_metaclass(ComponentizedBase)
+class Componentized(Component):
```
mamachanko: Not a fan of `Componentized`, but will not complain I can think of something better. On the other hand it goes well together with `Component`.

lymph/core/components.py
emulbreh: I should have added a WIP comment, this code will go away.

lymph/core/container.py
```diff
@@ -56,14 +57,18 @@ def __init__(self, ip='127.0.0.1', port=None, registry=None, logger=None, events
...
-        self.install(DefaultInterface, interface_name='lymph')
+        self.add_component(registry)
         registry.install(self)
         if events:
+            self.add_component(events)
             events.install(self)
 
+        self.server = self.install(self.server_cls, ip=ip, port=port)
+        self.monitor = self.install(MonitorPusher, endpoint=monitor_endpoint, interval=5)
+        self.install_interface(DefaultInterface, name='lymph')
```
mamachanko: The way components get added to and installed on each other is a little confusing. Throwing in `install_interface` doesn't make things easier to grok.

docs/api/components_api.rst
```diff
@@ -0,0 +1,49 @@
...
+    
+        Adds `component`.
+    
+    .. attribute:: metrics
+    
+        A :class:`Metrics` instance that aggregates the metrics from all added components.
+    
+    .. method:: on_start()
+    
+        Calls `on_start()` on all added components.
```
mamachanko: Shouldn't it be instead?

_"Should call `on_start()` on all added components."_

lymph/core/monitoring/pusher.py
```diff
@@ -0,0 +1,57 @@
...
+        self.loop_greenlet.kill()
+
+    def loop(self):
+        last_stats = time.monotonic()
+        while True:
+            gevent.sleep(self.interval)
+            dt = time.monotonic() - last_stats
+            series = list(self.aggregate)
+            stats = {
+                'dt': dt,
```
mamachanko: delta?

docs/api/components_api.rst
emulbreh: It *does* call `on_start()`, i.e. it's not an interface but a base class.

lymph/core/rpc.py
```diff
@@ -217,13 +220,7 @@ def _recv_loop(self):
...
-                'requests': dict(self.request_counts)
-            },
-            'connections': [c.stats() for c in self.connections.values()],
-        }
-        self.request_counts.clear()
-        return stats
+    def raw_metrics(self):
+        yield 'rpc.connection_count', len(self.connections), {}
+        for subject, count in self.request_counts.items():
+            yield 'rpc.request_count', count, {'subject': subject}
```
mamachanko: removed `self.request_counts.clear()` on purpose?

docs/api/components_api.rst
mamachanko: understood

lymph/core/monitoring/pusher.py
emulbreh: Then it'd have to be `delta_t` (or time_delta). This is just keeping the data we had in the stats before, but I think we can remove it completely now.

lymph/core/rpc.py
emulbreh: yes, we should monitor absolutes instead of rates.

lymph/core/components.py
manisenkov: `Compound` or `Composite`? Just came into my mind

mamachanko:
heavy stuff! nice.

lgtm

PR #148, opened 2015-03-27
Better errors during quickstart example

Drahflow:


lymph/config.py
```diff
@@ -35,10 +35,16 @@ def setdefault(self, key, default):
 
     def create_instance(self, key, default_class=None, **kwargs):
         instance_config = self.get(key, {})
-        return self._create_instance(instance_config, default_class=default_class, **kwargs)
+        return self._create_instance(key, instance_config, default_class=default_class, **kwargs)
+
+    def _create_instance(self, instance_key, instance_config, default_class=None, **kwargs):
```
emulbreh: get_instance() and create_instance() use `key`, let's keep this consistent.

lymph/config.py
```diff
@@ -35,10 +35,16 @@ def setdefault(self, key, default):
...
+        if instance_config is None:
+            raise ValueError("no config available for %r" % instance_key)
 
-    def _create_instance(self, instance_config, default_class=None, **kwargs):
         clspath = instance_config.get('class', default_class)
+        if clspath is None:
+            raise ValueError("no config available for %r (or no class configured)" % instance_key)
```
mouadino: why not just "no class configured %r" ? since "no config available for %r" error is already checked before ?

lymph/config.py
```diff
@@ -35,10 +35,16 @@ def setdefault(self, key, default):
 
     def create_instance(self, key, default_class=None, **kwargs):
         instance_config = self.get(key, {})
-        return self._create_instance(instance_config, default_class=default_class, **kwargs)
+        return self._create_instance(key, instance_config, default_class=default_class, **kwargs)
+
+    def _create_instance(self, instance_key, instance_config, default_class=None, **kwargs):
+        if instance_config is None:
+            raise ValueError("no config available for %r" % instance_key)
```
emulbreh: I'd rather raise a custom ConfigurationError here, since we don't report a problem with the value of the argument but with the configuration.

lymph/config.py
Drahflow: I did this because create_instance passes in {} if no configuration exists.

emulbreh:
+1

PR #149, opened 2015-03-27
Guide the newbee user to the included help

Drahflow:
Before this, a plain `lymph` would just show

    Usage: lymph [options] <command> [<args>...]

which leaves no clue about how to use `lymph help`


mouadino:
@Drahflow can you squash your commits pleases ? a commit that introduce a broken state is a useless  commit IMHO :)

mouadino:
LGTM

PR #150, opened 2015-03-27
Prevent KeyError during interpreter shutdown

emulbreh:


mouadino:
LGTM

Drahflow:
I wouldn't be so sure about this change. If some code actually *used* threading before this get's called, I'd expect fun and possibly hard to understand bugs to ensue...

emulbreh:
I assumed that pkg_resources (which we have no control over) was importing threading, but it turns out it was logging.

PR #152, opened 2015-03-30
Add an event mocking test case

mouadino:
This way we can easily mock events like we do with RPC in our tests.


dushyant88:
+1

mamachanko:
+1

PR #153, opened 2015-03-31
Add a way to poll metrics from an interface

mouadino:
Since all metrics are send to one 0mq subscriber endpoint it's harder
to get metrics from a given interface, by using get_metrics() RPC function
users can have a quick look on a given interface metrics.

This just a first step, since get_metrics() call will return results from
different instances each time it's called in case we are running different
service instance, so for now this should only be used for debugging purposes.


kashopi:
:+1:

PR #155, opened 2015-03-31
Add usefull metrics to proxy

mouadino:
This will give us statistics over proxies e.g. timeout and
exception counts.


manisenkov:
+1

PR #156, opened 2015-03-31
Container removal

emulbreh:


lymph/core/components.py
```diff
@@ -1,7 +1,13 @@
+import sys
+import gevent
```
netme: An empty line is missing

lymph/core/container.py
```diff
@@ -73,7 +66,8 @@ def __init__(self, ip='127.0.0.1', port=None, registry=None, events=None, node_e
 
         self.monitor = self.install(MonitorPusher, aggregator=self.metrics_aggregator, endpoint=self.monitor_endpoint, interval=5)
 
-        self.server = self.install(self.server_cls, ip=ip, port=port)
+        self.add_component(rpc)
+        rpc.request_handler = self.handle_request
```
mouadino: This is weird, why request_handler was moved here ?

lymph/core/container.py
```diff
@@ -157,12 +143,14 @@ def start(self, register=True):
         self.on_start()
         self.metrics_aggregator.add_tags(identity=self.identity)
 
+        self.instance = ServiceInstance(**self.get_instance_description())
```
mouadino: Why this is an instance variable of container ?

lymph/core/rpc.py
```diff
@@ -46,7 +56,7 @@ def identity(self):
 
     def _bind(self, max_retries=2, retry_delay=0):
         if self.bound:
-            raise TypeError('this container is already bound (endpoint=%s)', self.endpoint)
+            raise TypeError('already bound (endpoint=%s)', self.endpoint)
```
mouadino: s/TypeError/RuntimeError ?

lymph/core/rpc.py
```diff
@@ -109,33 +119,58 @@ def _close_sockets(self):
...
+        logger.debug('-> %s to %s', msg, endpoint)
         connection.on_send(msg)
 
-    def send_request(self, address, subject, body, headers=None):
+    def prepare_headers(self, headers):
+        headers = headers or {}
+        headers.setdefault('trace_id', trace.get_id())
+        return headers
+
+    def _pick_endpoint(self, service):
```
mouadino: I would like maybe in the future to have this outside this class and injected from argument, e.g.
``ZmqRPCServer(..., balancing_strategy='random')`` or 'round-robin' ... .

lymph/core/rpc.py
```diff
@@ -109,33 +119,58 @@ def _close_sockets(self):
...
+            try:
+                connection = self.connections[instance.endpoint]
+            except KeyError:
+                choices.append(instance.endpoint)
+                continue
+            if connection.is_alive():
+                choices.append(instance.endpoint)
+        if not choices:
+            raise NotConnected('Not connected to %s' % service.name)
+        return random.choice(choices)
```
mouadino: I would like maybe in the future to have this injected from the argument, e.g.
``ZmqRPCServer(..., balancing_strategy='random')`` or 'round-robin' ... .

lymph/core/container.py
emulbreh: The request handler needs the installed interfaces, and I don't want the rpc server to depend on container.

lymph/core/container.py
emulbreh: It's a description of the instance that the container runs. Not strictly necessary to keep in the instance, but it's supposed to be static during the lifetime of a container. 

I want to refactor this to allow for different descriptions for different interfaces later.

lymph/core/rpc.py
emulbreh: Yes, that would be nice.

lymph/core/rpc.py
```diff
@@ -45,8 +55,7 @@ def identity(self):
             return hashlib.md5(self.endpoint.encode('utf-8')).hexdigest()
 
     def _bind(self, max_retries=2, retry_delay=0):
-        if self.bound:
-            raise TypeError('this container is already bound (endpoint=%s)', self.endpoint)
+        assert not self.bound, 'already bound (endpoint=%s)' % self.endpoint
```
mouadino: An assert :( it's a RuntimeError not an AssertionError

emulbreh: It's a "this should never happen" check, i.e. a programming error.

mouadino:
We should not block this for too long since it change quite a few things.

LGTM

PR #157, opened 2015-03-31
Gpool double acquire fix

emulbreh:


mouadino:
@emulbreh why change the package name ?

emulbreh:
Because it's not a contrib package. It's used by core.

lymph/utils/gpool.py
```diff
@@ -15,9 +15,7 @@ class NonBlockingPool(Pool):
     When the ``timeout`` is not given or set to None the pool will reject
     immediately without waiting when pool size reach max size.
 
-    In case ``size`` is None this will create an unbound pool, this was
```
mouadino: Why ?

mouadino:
LGTM

lymph/utils/gpool.py
emulbreh: Because it's a misleading comment (it's not true at the moment), unbounded pools are the default.

PR #158, opened 2015-04-07
Added basic dead letter exchange based message delays

emulbreh:


lymph/events/kombu.py
```diff
@@ -130,7 +132,32 @@ def _get_connection(self):
         with kombu.pools.connections[self.connection].acquire(block=True) as conn:
             yield conn
 
-    def emit(self, event):
+    def _get_waiting_queue(self, conn, event_type, delay):
+        delay_ms = int(1000 * delay)
+        queue_name = '%s-wait_%s' % (event_type, delay_ms)
```
kpanic: I am wondering if there might be a "clash" between to queue names that have `anevent-wait_3000` if you would have two of them for the same even and delay time?

kpanic: Probably no since the Exchange name will be different?

lymph/tests/integration/test_kombu_events.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_emit(self):
+        self.lymph_client.emit('foo', {})
+        self.assert_eventually_true(self.received_check(1), timeout=2)
+        self.assertEqual(self.the_interface.collected_events[0].evt_type, 'foo')
+
+    def test_delayed_emit(self):
+        self.lymph_client.emit('foo', {}, delay=.5)
+        self.assert_temporarily_true(self.received_check(0), timeout=.2)
+        self.assert_eventually_true(self.received_check(1), timeout=.5)
```
mouadino: Not sure how stable this will be ? expecting the event to be receive exactly after .5  seconds ! maybe add some margin ?

lymph/tests/integration/test_kombu_events.py
```diff
@@ -0,0 +1,71 @@
...
+        return check
+
+    def test_emit(self):
+        self.lymph_client.emit('foo', {})
+        self.assert_eventually_true(self.received_check(1), timeout=2)
+        self.assertEqual(self.the_interface.collected_events[0].evt_type, 'foo')
+
+    def test_delayed_emit(self):
+        self.lymph_client.emit('foo', {}, delay=.5)
+        self.assert_temporarily_true(self.received_check(0), timeout=.2)
```
mouadino: IMHO I don't this is worth it, named are hard to understand and I believe this can be easily written with just simple sleep() then assertEqual. waiting for extra .2 or .5 or event 1 second should not matter compared to readability. ofc Non-Blocking :)

lymph/events/kombu.py
```diff
@@ -130,7 +132,32 @@ def _get_connection(self):
...
+        queue(conn).bind_to(exchange=self.waiting_exchange, routing_key=queue_name)
+        self.waiting_queues[queue_name] = queue
+        return queue
+
+    def emit(self, event, delay=0):
         with self._get_connection() as conn:
             producer = conn.Producer(serializer=self.serializer)
-            producer.publish(event.serialize(), routing_key=event.evt_type, exchange=self.exchange, declare=[self.exchange])
+            if delay:
+                queue = self._get_waiting_queue(conn, event.evt_type, delay)
```
mouadino: s/_get_waiting_queue/_create_waiting_queue/ ?

lymph/events/kombu.py
kpanic: got it @emulbreh thank you ;)

lymph/tests/integration/test_kombu_events.py
emulbreh: there's a 200ms wait before this check, so we have a margin of 200ms. Is that too optimistic?

lymph/events/kombu.py
emulbreh: it's basically get_or_create, with `get` (hopefully) being more common than `create`.

lymph/events/kombu.py
mamachanko: so is it because of the exchange name?

lymph/testing/__init__.py
```diff
@@ -259,3 +260,29 @@ def setUp(self):
...
+        try:
+            self.wait_for(condition, timeout=timeout)
+        except gevent.Timeout:
+            self.fail(message)
+
+    def assert_temporarily_true(self, condition, message=None, timeout=2):
+        try:
+            with gevent.Timeout(timeout):
+                while condition():
+                    gevent.sleep(0)  # yield
```
mamachanko: Shouldn't this be `self.wait_for(condition, timeout=timeout)`?

lymph/tests/integration/test_kombu_events.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def received_check(self, n):
+        def check():
+            return len(self.the_interface.collected_events) == n
+        return check
+
+    def test_emit(self):
+        self.lymph_client.emit('foo', {})
+        self.assert_eventually_true(self.received_check(1), timeout=2)
+        self.assertEqual(self.the_interface.collected_events[0].evt_type, 'foo')
```
mamachanko: I like it.

lymph/events/kombu.py
emulbreh: If two events have the same event type and delay, they have the same event type and delay. So they can go to the queue that handles all events with this event type and delay.

lymph/testing/__init__.py
mamachanko: I get it. The condition would have to be negated. Double-negated even.

mamachanko:
lgtm

PR #160, opened 2015-04-07
Refactor metrics and add new out of the box metrics classes

mouadino:
Refactor metrics and add new out of the box metrics classes
    
Split Metrics class into an Aggregator and RawMetric/Counter/NamespaceCounter
metric data holders, replace the tree-like behavior of Metrics class for each
component to a simple _get_metrics method in each component and python super(...).

### DIFF
--- a/docs/api/components_api.rst
+++ b/docs/api/components_api.rst
@@ -4,26 +4,27 @@
 Components API
 ==============
 
-Components are objects that depend on a running service container. They are 
-embedded in :class:`Componentized` objects. 
-Since Componentized objects themselves are components, they form a tree of 
-:class:`Component` instances with the container as the root.
+Components are objects that depend on a running service container. They are
+embedded in :class:`Componentized` objects.
+Since Componentized objects themselves are components, they form a tree of
+:class:`Component` instances with the container as the root. An example
+of a Component is :class:`lymph.core.interfaces.Interface`.
 
 
 .. class:: Component
 
-    .. attribute:: metrics
-    
-        A :class:`Metrics` instance that will be included in the monitoring data
-        of the conatainer.
-
     .. method:: on_start()
-    
+
         Called when the container is started.
 
     .. method:: on_stop()
-    
+
         Called when the container is stopped.
+
+    .. method:: _get_metrics()
+
+        Is being called to get metrics. Returns an iterable or yields
+        values of type :class:`lymph.core.monitoring.metrics.RawMetric`.
 
 
 .. class:: Componentized()
@@ -31,19 +32,15 @@ Since Componentized objects themselves are components, they form a tree of
     A collection of components; itself a component.
 
     .. method:: add_component(component)
-    
+
         :param component: :class:`Component`
-    
+
         Adds `component`.
-    
-    .. attribute:: metrics
-    
-        A :class:`Metrics` instance that aggregates the metrics from all added components.
-    
+
     .. method:: on_start()
-    
+
         Calls `on_start()` on all added components.
-    
+
     .. method:: on_stop()
 
         Calls `on_stop()` on all added components.
--- a/docs/api/metrics_api.rst
+++ b/docs/api/metrics_api.rst
@@ -1,0 +1,37 @@
+.. currentmodule:: lymph.core.monitoring.metrics
+
+
+Metrics API
+===========
+
+
+.. class:: RawMetric(name, value, tags=None)
+
+   Raw metric that represent an arbitrary value a.k.a Gauge Metric.
+
+   .. method:: __iter__()
+
+      Yield metric values as a tuple in the form `(name, value, tags)`.
+
+
+.. class:: Counter(name, tags=None)
+
+    A counter is a cumulative metric that represents a single numerical
+    value that only ever goes up. A counter is typically used to count
+    requests served, tasks completed, errors occurred, etc.
+
+    .. method:: __iadd__(new_value=1)
+
+      Increment counter value.
+
+
+.. class:: TaggedCounter(name, tags=None)
+
+    A tagged counter is a container metric that represents multiple
+    counters per tags. A tagged counter is typically used to track a group
+    of counters as one e.g. request served per function name, errors ocurred
+    per exception name, etc.
+
+    .. method:: incr(_value=1, **tags)
+
+      Increment given counter ``type`` by ``new_value``.
--- a/lymph/core/components.py
+++ b/lymph/core/components.py
@@ -1,17 +1,16 @@
 import six
 
-from lymph.core.monitoring.metrics import Metrics
-
 
 class Component(object):
-    def __init__(self):
-        self.metrics = Metrics()
 
     def on_start(self):
         pass
 
     def on_stop(self, **kwargs):
         pass
+
+    def _get_metrics(self):
+        return []
 
 
 class Declaration(object):
@@ -57,7 +56,6 @@ class Componentized(Component):
 
     def add_component(self, component):
         self.__all_components.append(component)
-        self.metrics.add(component.metrics)
 
     def install(self, factory, **kwargs):
         if factory in self._declared_components:
@@ -77,3 +75,10 @@ class Componentized(Component):
     def on_stop(self, **kwargs):
         for component in reversed(self.__all_components):
             component.on_stop(**kwargs)
+
+    def _get_metrics(self):
+        for metric in super(Componentized, self)._get_metrics():
+            yield metric
+        for component in self.__all_components:
+            for metric in component._get_metrics():
+                yield metric
--- a/lymph/core/container.py
+++ b/lymph/core/container.py
@@ -12,7 +12,9 @@ import six
 from lymph.exceptions import RegistrationFailure, SocketNotCreated
 from lymph.core.components import Componentized
 from lymph.core.events import Event
+from lymph.core.monitoring import metrics
 from lymph.core.monitoring.pusher import MonitorPusher
+from lymph.core.monitoring.aggregator import Aggregator
 from lymph.core.services import ServiceInstance, Service
 from lymph.core.rpc import ZmqRPCServer
 from lymph.core.interfaces import DefaultInterface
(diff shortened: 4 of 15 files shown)


docs/api/components_api.rst
```diff
@@ -4,46 +4,43 @@
...
 ==============
 
-Components are objects that depend on a running service container. They are 
-embedded in :class:`Componentized` objects. 
-Since Componentized objects themselves are components, they form a tree of 
-:class:`Component` instances with the container as the root.
+Components are objects that depend on a running service container. They are
+embedded in :class:`Componentized` objects.
+Since Componentized objects themselves are components, they form a tree of
+:class:`Component` instances with the container as the root. an example
```
mamachanko: _An_

docs/api/components_api.rst
```diff
@@ -4,46 +4,43 @@
...
-    
+
         Called when the container is started.
 
     .. method:: on_stop()
-    
+
         Called when the container is stopped.
 
+    .. method: _get_metrics()
```
mamachanko: double colon

docs/api/components_api.rst
```diff
@@ -4,46 +4,43 @@
...
+
+        Called to get metrics should return an iterable or yield
+        values of type :class:`lymph.core.monitoring.metrics.RawMetric`.
```
mamachanko: Is being called to get metrics. Returns an iterable or yields values of type ...

(suggestion)

docs/api/metrics_api.rst
```diff
@@ -0,0 +1,41 @@
+.. currentmodule:: lymph.core.monitoring.metrics
+
+
+Metrics API
+===========
+
+
+.. class:: RawMetric(name, value, tags=None)
+
+   Raw metric that represent an arbitararily value a.k.a Gauge Metric.
```
mamachanko: arbitrary

mamachanko: Is GaugeMetric the basic metric or is it ex pluribus unum?

docs/api/metrics_api.rst
```diff
@@ -0,0 +1,41 @@
...
+    requests served, tasks completed, errors occurred, etc.
+
+    .. method:: __iadd__(new_value=1)
+
+      Increment counter value.
+
+
+.. class:: NamespaceCounter(root_ns, tags=None)
+
+    A namespace counter is container metric that represents multiple
```
mamachanko: is _a_ container

lymph/core/components.py
```diff
@@ -13,6 +9,9 @@ def on_start(self):
     def on_stop(self, **kwargs):
         pass
 
+    def _get_metrics(self):
```
mamachanko: why "private" if it's piublicly documented?

lymph/core/container.py
```diff
@@ -202,3 +212,6 @@ def emit_event(self, event_type, payload, headers=None):
 
     def send_request(self, address, subject, body, headers=None):
         return self.server.send_request(address, subject, body, headers=None)
+
+    def _get_metrics(self):
```
mamachanko: what if called it `metrics` and turned it into a property? there's already `endpoint`, `service_types` and `identity`. I'd like it.

lymph/core/interfaces.py
```diff
@@ -75,7 +73,7 @@ def _call(self, __name, **kwargs):
             return channel.get(timeout=self._timeout).body
         except RemoteError as e:
             error_type = str(e.__class__)
-            self.exception_counts[e.__class__.__name__] += 1
+            self.exception_counts.incr(e.__class__.__name__)
```
mamachanko: Just checking whether it looks good:

``` python
self.exception_counts(e.__class__.__name__) += 1
```

docs/api/metrics_api.rst
mouadino: It's the basic metric, so all other Metrics inherit from this one, not sure how to represent that in sphinx doc :(

lymph/core/rpc.py
```diff
@@ -32,14 +33,12 @@ def __init__(self, container, ip='127.0.0.1', port=None):
         self.zctx = zmq.Context.instance()
         self.endpoint = None
         self.bound = False
-        self.request_counts = collections.Counter()
+        self.request_metric = metrics.NamespaceCounter('rpc')
```
mamachanko: why rename?

lymph/core/components.py
mouadino: Leading underscore mean internal since as you know there is no private in Python, but the question remain is internal to what ? i.e. what is external ? :)

Enough philosophy :P as for why the leading underscore that because we don't want to forbid using get_metrics as an RPC function name which is what we do here https://github.com/deliveryhero/lymph/blob/master/lymph/core/interfaces.py#L185

lymph/core/container.py
mouadino: I don't like it, because properties are hard to reason about ! is it computed each time or memoized !? pure OOP you never access attribute directly in your object you always ask object for something by sending a message or calling a function, this is IMHO easy to reason about.

BTW any reason why you want this a property ?

lymph/core/interfaces.py
mouadino: what do you expect calling/instantiating ``self.exception_counts`` to return !?

lymph/core/rpc.py
mouadino: because for me it read better, since with this change we are dealing with metrics and not abstract object like counts.

lymph/tests/monitoring/test_aggregator.py
```diff
@@ -0,0 +1,48 @@
...
+class AggregatorTestCase(object):
+
+    def test_aggregator_one_component(self):
+        aggr = Aggregator(name='test')
+
+        aggr.register_component(DummyComponentOne())
+
+        self.assertEqual(list(aggr.get_metrics()), [('dummy', 'one', {'name': 'test'})])
+
+    def test_aggregator_multiple_component(self):
```
mamachanko: components

lymph/tests/monitoring/test_aggregator.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def test_aggregator_add_tags(self):
+        aggr = Aggregator(name='test')
+        aggr.register_component(DummyComponentOne())
+
+        aggr.add_tags(origin='localhost')
+
+        self.assertEqual(list(aggr.get_metrics()), [
+            ('dummy', 'one', {'name': 'test', 'origin': 'localhost'})
+        ])
```
mamachanko: adding multiple tags?

mouadino: don't get your comment !

lymph/tests/monitoring/test_metrics.py
```diff
@@ -0,0 +1,95 @@
+import operator
+import unittest
+
+from lymph.core.monitoring import metrics
+
+
+class RawMetricsTest(unittest.TestCase):
+
+    def test_repr(self):
+        self.raw = metrics.RawMetric('raw', 'foobar', {})
```
mamachanko: why make it an instance attirbute?

lymph/tests/monitoring/test_metrics.py
```diff
@@ -0,0 +1,95 @@
...
+
+        self.assertEqual(list(self.raw), [
+            ('raw', 'foobar', {'origin': 'localhost', 'time': 'now', 'name': 'raw'}),
+        ])
+
+
+class CounterMetricsTest(unittest.TestCase):
+
+    def test_get(self):
+        self.counter = metrics.Counter('requests')
```
mamachanko: same here. why is it an instance attribute?

lymph/tests/monitoring/test_metrics.py
```diff
@@ -0,0 +1,95 @@
...
+
+        self.assertEqual(list(self.counter), [('requests', 0, {})])
```
mamachanko: why make it a list?

docs/api/metrics_api.rst
mamachanko: Isn't RawMetric the basic metric?

lymph/core/components.py
mamachanko: thanks

lymph/core/container.py
mamachanko: no, not in particular. thanks.

lymph/core/interfaces.py
mamachanko: Would expect to return the counter in the namespace. But yeah... forget about it.

lymph/tests/monitoring/test_metrics.py
mouadino: b/c that's how you access metrics raw data (i.e. ``__iter__``).

lymph/core/rpc.py
mamachanko: what about these two counts then? https://github.com/deliveryhero/lymph/pull/160/files#diff-8a6086eb1e8566975ef71c7b6ca2051fR67

lymph/tests/monitoring/test_aggregator.py
mamachanko: what about testing the case of adding multiple tags?

lymph/tests/monitoring/test_metrics.py
mamachanko: I get it.

lymph/core/rpc.py
mouadino: it doesn't matter much for me, so I am reverting to old name :)

lymph/core/container.py
```diff
@@ -57,16 +59,23 @@ def __init__(self, ip='127.0.0.1', port=None, registry=None, events=None, node_e
         self.installed_plugins = []
 
         self.debug = debug
+        self.monitor_endpoint = monitor_endpoint
+
+        self.metrics_aggregator = Aggregator(service=self.service_name, host=self.fqdn)
+        self.metrics_aggregator.register_component(self)
+
+        if self.service_registry:
```
emulbreh: why `if` here? (for symmetry: I'd rather make event_system required)

lymph/core/monitoring/metrics.py
```diff
@@ -1,26 +1,59 @@
...
+
+    def __iadd__(self, value):
+        self._value += value
+        return self
+
+    def reset(self):
+        self._value = 0
+
+
+class NamespaceCounter(RawMetric):
```
emulbreh: `NamespacedCounter` would be better since we're not counting namespaces.

mamachanko:
+1 from me

lymph/core/monitoring/metrics.py
```diff
@@ -1,26 +1,48 @@
...
+
+    def __init__(self, name, tags=None):
+        super(Counter, self).__init__(name, 0, tags)
+
+    def __iadd__(self, value):
+        self._value += value
+        return self
+
+
+class TaggedCounter(RawMetric):
```
emulbreh: I still think this should be the Counter class :)

docs/api/metrics_api.rst
```diff
@@ -0,0 +1,41 @@
...
+
+Metrics API
+===========
+
+
+.. class:: RawMetric(name, value, tags=None)
+
+   Raw metric that represent an arbitrary value a.k.a Gauge Metric.
+
+   .. method:: add_tags(**tags)
```
emulbreh: This method doesn't exist anymore.

emulbreh:
I'm still not convinced that this is the better api, but this isn't blocking. +1 (except for the add_tags() docs)

lymph/core/monitoring/metrics.py
mouadino: I still think that splitting them is the natural thing to do since use case wise you either use one behaviour or the another but not both, but we will see what the future will bring :)

mouadino:
For me the super() + _get_metrics() make the tree logic more explicit, but yes let's see when we come back to this how we fill about it :)

PR #161, opened 2015-04-08
Update RequsetHandler

lyudmildrx:
Make use of werkzeug request.get_data in json method. Wekrzeug
handles charset and does caching for us.

As long as method verifies mimetype is json the method is
safe from messing with form data.

### DIFF
--- a/lymph/web/handlers.py
+++ b/lymph/web/handlers.py
@@ -1,4 +1,3 @@
-import codecs
 import json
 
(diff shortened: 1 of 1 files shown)


emulbreh:
+1

mouadino:
Bring some memories https://github.com/deliveryhero/lymph/pull/94#discussion-diff-23831041R26 :)

PR #162, opened 2015-04-10
Make registry and event optional

mouadino:
This will make using ServiceContainer in our tests much easier so that
we can take advantage of EventMockTestCase.


mamachanko:
lgtm

PR #164, opened 2015-04-13
Add mock helpers to make working checking mock calls easier

mouadino:
First step toward better testing suite.

### DIFF
--- a/lymph/tests/test_mock_helpers.py
+++ b/lymph/tests/test_mock_helpers.py
@@ -1,0 +1,58 @@
+import unittest
+
+import mock
+
+from lymph.testing.mock_helpers import MockMixins
+
+
+class DummyTestCase(unittest.TestCase, MockMixins):
+
+    def runTest(self):
+        pass
+
+
+class RPCMockHelperTests(unittest.TestCase):
+
+    def setUp(self):
+        self.dummy_case = DummyTestCase()
+
+    def test_single_call_match(self):
+        self.dummy_case._assert_equal_calls(
+            [mock.call('func', 1, foo='bar')],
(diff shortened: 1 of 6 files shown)


bilunyk:
:+1: hamcrest is pretty sweet

mouadino:
@bilunyk It's quite good, but writting custom matchers is hard to do specially getting error description right http://pyhamcrest.readthedocs.org/en/release-1.8/custom_matchers/ or maybe I am just too stupid to figure out how to do it properly :)

lymph/tests/test_mock_helpers.py
```diff
@@ -0,0 +1,58 @@
...
+        self.dummy_case = DummyTestCase()
+
+    def test_single_call_match(self):
+        self.dummy_case._assert_equal_calls(
+            [mock.call('func', 1, foo='bar')],
+            [mock.call('func', 1, foo='bar')]
+        )
+
+    def test_single_call_args_mismatch(self):
+        with self.assertRaises(AssertionError):
```
mamachanko: Did you consider asserting with `self.assertRaisesRegexp` ot see whether the correct message is included?

mouadino: Yes, I did but the error message are a work in progress, for keyword name you get a better error message than before, but not good enough, this is where I was thinking hamcrest matchers may be useful but then I gave up on them for now, but definitely when we have good enough message we can check them in our tests.

mamachanko: understood

mouadino: Correction: I mean hamcrest **custom** matchers, the idea is still to use the given matchers ofc :)

PR #165, opened 2015-04-14
Remove 'class' from instance_config dict only if key exists

promulo:
This will avoid raising KeyError if 'class' is not a key of instance_config dict.

### DIFF
--- a/lymph/config.py
+++ b/lymph/config.py
@@ -51,7 +51,7 @@ class ConfigObject(collections.Mapping):
             return cls.from_config(instance_config, **kwargs)
         else:
             instance_config = copy.deepcopy(dict(instance_config))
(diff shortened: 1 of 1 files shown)


mouadino:
I fail to see which case it raise an error can you past here the config used ?

promulo:
It happened to me when I tried to use config.create_instance('urbanairship', default_class='urbanairship.Airship') with this config file:

```
interfaces:
    push_mobile:
        class: dhh.push_mobile.interfaces:PushMobile
        urbanairship:
            key: my_key
            secret: my_secret
```

mouadino:
hmm I see, I so wish I could nuke lymph config.py, but for now :+1:

PR #166, opened 2015-04-15
Send notifications to lymph IRC channel

promulo:


Menda:
There's a chat channel about Lymph? Didn't know that

promulo:
Yes. It's in irc.freenode.net network, #lymph :)

kpanic:
@promulo should we use http://docs.travis-ci.com/user/notifications/#IRC-notification from here
`use_notice: true` ? What do you think?

promulo:
you're right, @kpanic I've updated the PR, thanks :)

Menda:
There's a chat channel about Lymph? Didn't know that

:+1:

PR #168, opened 2015-04-16
Add a new method assert_any_rpc_calls

mouadino:
This can be useful when we only one to check for a subset of mock.calls, this
method like it's sibling try to generate the best error message by guessing
what user want to test for.


manisenkov:
LGTM

PR #173, opened 2015-04-21
Made python -m lymph.cli.main work

emulbreh:


eguven:
:+1:

mouadino:
Better yet, define ``__main__.py`` inside lymph/cli and use "python -m lymph.cli ..."

PR #174, opened 2015-04-23
Change RpcMockTestCase to forward to the real service when no mock is su...

mouadino:
...pplied for the current rpc function


jacqueslh:
LGTM

PR #175, opened 2015-04-24
document tail printing to stderr

mamachanko:
Allows to forward tail output to a file:

``` python
$ lymph tail --level=DEBUG my_service > /tmp/my_service.log
```


mouadino:
Is there a problem with doing

    lymph tail --level=DEBUG my_service 2> /tmp/my_service.log

?

mouadino:
There is a reason why logging Stream default is to stderr which is buffering AFAIK, stderr is unbuffered while stdout is buffered (to line buffering if it's terminal) ...

mouadino:
LGTM

PR #180, opened 2015-05-08
Publish http_endpoint of WebServiceInterfaces

emulbreh:


eguven:
:+1:

lymph/core/container.py
```diff
@@ -128,14 +128,16 @@ def subscribe(self, handler, **kwargs):
     def unsubscribe(self, handler):
         self.event_system.unsubscribe(handler)
 
-    def get_instance_description(self):
-        return {
+    def get_instance_description(self, interface):
+        description = interface.get_description()
+        description.update({
```
mouadino: don't we have a priority problem here since some data may get overrited ? maybe a different namespace ?

    {
        'interface_description': description,
        'endpoint': ...,
        ...
    }

Just to not have any surprises :)

emulbreh: I'd rather keep the namespace flat for now, I don't expect there to be many proper use cases for extra keys outside of lymph core.

PR #182, opened 2015-05-11
get_empty_kwargs need to be override as doc[1] say

mouadino:
With this we can use HandledRule with Submount and EndpointPrefix and so on.

[1]: http://werkzeug.pocoo.org/docs/0.10/routing/#werkzeug.routing.Rule.empty


manisenkov:
:+1:

PR #189, opened 2015-05-21
Event broadcast

itakouna:
Event broadcast allows multiple event's consumers to consume the same event and carry out concurrent tasks when an event is fired. We need it for multiple web-socket servers.  
Example:
```
@lymph.event('uppercase_transform_finished', broadcast=True)
 def on_uppercase(self, event):

```

### DIFF
--- a/lymph/core/events.py
+++ b/lymph/core/events.py
@@ -43,7 +43,7 @@ class Event(object):
 
 class EventHandler(Component):
     def __init__(self, interface, func, event_types, sequential=False, queue_name=None, active=True, once=False, broadcast=False):
-        assert (not (once and broadcast)), ("Once and broadcast cannot be enabled at the same time")
+        assert not (once and broadcast), "Once and broadcast cannot be enabled at the same time"
         super(EventHandler, self).__init__()
         self.func = func
         self.event_types = event_types
--- a/lymph/core/events.py
+++ b/lymph/core/events.py
@@ -1,7 +1,6 @@
 import re
 import logging
 from uuid import uuid4
-
 from lymph.core.interfaces import Component
 from lymph.core import trace
 
@@ -43,8 +42,8 @@ class Event(object):
 
 
 class EventHandler(Component):
-    def __init__(self, interface, func, event_types, sequential=False,
-                 queue_name=None, active=True, once=False, broadcast=False):
+    def __init__(self, interface, func, event_types, sequential=False, queue_name=None, active=True, once=False, broadcast=False):
+        assert (not (once and broadcast)), ("Once and broadcast cannot be enabled at the same time")
         super(EventHandler, self).__init__()
         self.func = func
         self.event_types = event_types
@@ -117,4 +116,3 @@ class EventDispatcher(object):
                 handlers.add(handler)
                 handler(event)
         return bool(handlers)
-
--- a/lymph/core/events.py
+++ b/lymph/core/events.py
@@ -43,7 +43,8 @@ class Event(object):
 
 
 class EventHandler(Component):
-    def __init__(self, interface, func, event_types, sequential=False, queue_name=None, active=True, once=False):
+    def __init__(self, interface, func, event_types, sequential=False,
+                 queue_name=None, active=True, once=False, broadcast=False):
         super(EventHandler, self).__init__()
         self.func = func
         self.event_types = event_types
@@ -51,7 +52,7 @@ class EventHandler(Component):
         self.active = active
         self.interface = interface
         self.once = once
-        self.unique_key = str(uuid4()) if once else None
+        self.unique_key = str(uuid4()) if once or broadcast else None
         self._queue_name = queue_name or func.__name__
 
     @property
--- a/lymph/core/events.py
+++ b/lymph/core/events.py
@@ -52,6 +52,7 @@ class EventHandler(Component):
         self.active = active
         self.interface = interface
         self.once = once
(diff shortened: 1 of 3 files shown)


mamachanko:
But events are broadcast already, aren't they? However, only one instance of each subscriber gets to consume. That is, if two services were to subscribe one instance of each service would consume the event. What are you trying to solve?

itakouna:
We need it for multiple web-socket servers, so the event is consumed for all instances that subscribe for that event.

emulbreh:
The queue should be `exclusive`, otherwise they'll hang around for ever and accumulate messages that are never consumed. They also don't need to be durable.

mamachanko:
Ok, the web-socket server problem makes sense. However, how's that `broadcast` flag going to solve it? `once` and `broadcast` literally do the same thing. What you want is to have an event consumed by every instance of a consumer.

itakouna:
It tried with ```once```, but the queue is created for a subscriber and after one event is deleted, which makes the subscriber not receiving upcoming events. Could you explain to me what ```once``` else does!?

emulbreh:
`once` is for a different use-case (only receive the next event and then unsubscribe).

lymph/core/events.py
```diff
@@ -43,7 +43,8 @@ def serialize(self):
 
 
 class EventHandler(Component):
-    def __init__(self, interface, func, event_types, sequential=False, queue_name=None, active=True, once=False):
+    def __init__(self, interface, func, event_types, sequential=False,
+                 queue_name=None, active=True, once=False, broadcast=False):
         super(EventHandler, self).__init__()
```
emulbreh: Let's raise an exception if ``once and broadcast``. An assert would be good enough.

lymph/core/events.py
```diff
@@ -43,7 +42,8 @@ def serialize(self):
 
 
 class EventHandler(Component):
-    def __init__(self, interface, func, event_types, sequential=False, queue_name=None, active=True, once=False):
+    def __init__(self, interface, func, event_types, sequential=False, queue_name=None, active=True, once=False, broadcast=False):
+        assert (not (once and broadcast)), ("Once and broadcast cannot be enabled at the same time")
```
emulbreh: (no need for) (the parenthesis here).

lymph/tests/integration/test_kombu_events.py
```diff
@@ -27,8 +37,15 @@ def setUp(self):
         self.exchange_name = 'test-%s' % uuid.uuid4()
         self.discovery_hub = StaticServiceRegistryHub()
 
-        self.the_container, self.the_interface = self.create_container(TestInterface, 'test')
-        self.lymph_client = self.create_client()
+        self.the_container1, self.the_interface1 = self.create_container(TestInterface, 'test')
+        self.lymph_client1 = self.create_client()
+
+        self.the_container2, self.the_interface2 = self.create_container(TestEventBroadcastInterface, 'test')
+        self.lymph_client2 = self.create_client()
```
emulbreh: one client should be enough, no?

lymph/tests/integration/test_kombu_events.py
```diff
@@ -27,8 +37,15 @@ def setUp(self):
         self.exchange_name = 'test-%s' % uuid.uuid4()
         self.discovery_hub = StaticServiceRegistryHub()
 
-        self.the_container, self.the_interface = self.create_container(TestInterface, 'test')
-        self.lymph_client = self.create_client()
+        self.the_container1, self.the_interface1 = self.create_container(TestInterface, 'test')
+        self.lymph_client1 = self.create_client()
+
+        self.the_container2, self.the_interface2 = self.create_container(TestEventBroadcastInterface, 'test')
```
emulbreh: Let's name these interfaces instead of numbering them, they are different after all.

lymph/tests/integration/test_kombu_events.py
itakouna: I think we need at least two to test broadcast feature.

lymph/tests/integration/test_kombu_events.py
```diff
@@ -27,8 +37,15 @@ def setUp(self):
         self.exchange_name = 'test-%s' % uuid.uuid4()
         self.discovery_hub = StaticServiceRegistryHub()
 
-        self.the_container, self.the_interface = self.create_container(TestInterface, 'test')
-        self.lymph_client = self.create_client()
+        self.the_container_unicast, self.the_interface_unicast = self.create_container(TestInterface, 'test')
+        self.lymph_client_unicast = self.create_client()
```
emulbreh: You only need one client. It only acts as another service to send messages/events from, and you can reuse it in all your test methods.

lymph/tests/integration/test_kombu_events.py
```diff
@@ -27,9 +31,12 @@ def setUp(self):
         self.exchange_name = 'test-%s' % uuid.uuid4()
         self.discovery_hub = StaticServiceRegistryHub()
 
-        self.the_container, self.the_interface = self.create_container(TestInterface, 'test')
+        self.the_container_emitter, self.the_interface_emitter = self.create_container(TestInterface, 'test')
         self.lymph_client = self.create_client()
 
+        self.the_container_receiver, self.the_interface_receiver = self.create_container(TestInterface, 'test')
+        self.create_client()
```
emulbreh: This line doesn't do anything.

PR #193, opened 2015-05-25
Allow object level serialization to be defined

mouadino:


docs/topics/serialization.rst
```diff
@@ -27,3 +27,42 @@ can be deserialized. The wrapping has the following form:
 
 
 .. _msgpack: www.msgpack.org
+
+
+Object level serialization
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Object level serialization can be defined by implement ``__lymph_dump__`` method in classes subject to serialization.
```
mamachanko: ... by implementing ...

docs/topics/serialization.rst
```diff
@@ -27,3 +27,42 @@ can be deserialized. The wrapping has the following form:
...
+
+Using object level serialization can lead in some situation to a more consist code, example of one of those situations:
```
mamachanko: my suggestion:

_Object-level serialization can help to produce more concise code in certain situations, e.g.:_

Yes, I think it's "_object-level serialization_".

docs/topics/serialization.rst
```diff
@@ -27,3 +27,42 @@ can be deserialized. The wrapping has the following form:
...
+
+        def __lymph_dump__(self):
+            return {
+                'pid': self.pid,
+                'name': self.name,
+            }
+
+
+    class Node(lymph.Interface):
+
```
mamachanko: "..."

docs/topics/serialization.rst
```diff
@@ -27,3 +27,42 @@ can be deserialized. The wrapping has the following form:
...
+        def get_processes(self, service_type=None):
+            procs = []
+            for proc in self._processes:
+                if not service_type or proc.service_type == service_type:
+                    procs.append(proc)
+            return procs
+
+        @lymph.rpc()
+        def stop(self, service_type=None):
+            for proc in self.get_processes(service_type):
```
mamachanko: This call won't return the custom-serialized `Process`, correct?

emulbreh:
I like the idea, but I don't like the method name (PEP8: "Never invent such names; only use them as documented."). Proposals:

* Pick a single underscore name: `def _lymph_serialize(self)`
* Use a plain (fixed) method name and bless it: 
```python
  @lymph.serializers.bless
  def serialize(self): pass
```
* Use a class decorator
```python
@lymph.serializers.register('dump')
class Foo(object):
    def dump(self): pass
```
* Registration pattern:
```python
class Foo(object):
    def serialize(self): pass

lymph.serializers.register(Foo, dump='serialize')
```
My favorite is the last option.

mouadino:
I prefer duck typing :)

I don't like the word "serialize" since it hard to type as I mentioned in the spec.

As for why dunder b/c that how protocol are defined in python, so it's fell natural. I know that pep8 don't advice to use dunder names but I thought of ignoring it since the reason for the advice doesn't apply here that's also a reason why the lymph prefix is also there.

but if you want to follow pep8 nevertheless I can go with ``_lymph_dump_``, thought ?

docs/topics/serialization.rst
mouadino: sorry ?

docs/topics/serialization.rst
mouadino: done

docs/topics/serialization.rst
mouadino: it will return what it say it will return which is ``Process`` instances, serialization will be handled by lymph in this case.

mouadino:
@emulbreh @mamachanko changed to `_lymph_dump_` since as explained before, please check again :)

docs/topics/serialization.rst
mamachanko: aren't you skipping methods between `class Node` and `def get_processes` or is `self._processes` an attribute of `lymph.Interface`? anyway. non-blocking.

docs/topics/serialization.rst
```diff
@@ -27,3 +27,41 @@ can be deserialized. The wrapping has the following form:
 
 
 .. _msgpack: www.msgpack.org
+
+
+Object level serialization
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Object level serialization can be defined by implementing ``_lymph_dump_`` method in classes subject to serialization.
```
mamachanko: Isn't it `__lymph_dump__`?

mouadino: Fixed

mamachanko:
looks good to me

PR #194, opened 2015-05-26
Request SIGKILL on parent death if prctl is available

emulbreh:
Not sure whether this is a good idea.

### DIFF
--- a/lymph/cli/service.py
+++ b/lymph/cli/service.py
@@ -8,6 +8,11 @@ import gevent
 from setproctitle import setproctitle
 import six
 
+try:
+    import prctl
+except ImportError:
+    prctl = None
+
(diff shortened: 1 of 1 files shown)


lymph/cli/service.py
```diff
@@ -56,6 +61,9 @@ class InstanceCommand(Command):
     def run(self):
         debug = self.args.get('--debug')
 
+        if prctl and os.environ.get('LYMPH_NODE'):
+            prctl.set_pdeathsig(signal.SIGKILL)
```
Drahflow: Why not SIGINT, so we could ensure a possibly still running request is being completed before the instance is shut down? Or do we expect all clients to handle spontaneous failures gracefully (which would make for a more resilient system probably, but would mean more work overall)?

PR #195, opened 2015-05-26
add dynamic spacing to list command

mamachanko:


MislavStipetic:
+1

lymph/cli/list.py
```diff
@@ -14,10 +14,14 @@ class ListCommand(Command):
     needs_config = False
 
     def run(self):
+        command_names = get_command_classes().keys()
+        max_command_name = max(command_names, key=len)
+        description_offset = len(max_command_name) + 2
```
MislavStipetic: maybe you can fold these two into
description_offset = max([(len(name) + 2) for name in command_names])
but I'm being pedantic :)

PR #198, opened 2015-06-04
Forward exception information on unexpected errors

Drahflow:
... because it's nice when you don't have to hunt for the original Nack source.

### DIFF
--- a/lymph/core/channels.py
+++ b/lymph/core/channels.py
@@ -23,7 +23,7 @@ class RequestChannel(Channel):
         try:
             msg = self.queue.get(timeout=timeout)
             if msg.type == Message.NACK:
-                raise Nack(self.request)
+                raise Nack(self.request, msg.body)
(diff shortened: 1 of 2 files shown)


jacqueslh:
LGTM

Menda:
It's fine :+1:

emulbreh:
NACK intentionally doesn't contain a traceback or other error details, since you can't recover from this type of error. Why do you want it?

Drahflow:
Traceback information is not for recovering from error, but to aid in debugging. It is helpful during development when new errors happen and also in production as one does not have to hunt through the error logs of multiple services and hope to correlate multiple errors into one narrative.

Given that this only concerns unrecoverable errors, which we don't expect too often on production, I don't see much of a downside of sending this extra info.

Drahflow:
I made #203 so the proposed design is actually implemented and grepping the logfiles useful.

PR #200, opened 2015-06-08
Added support for environment variable substitution in config

emulbreh:
Syntax borrowed from [Circus](http://circus.readthedocs.org/en/latest/)

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -21,6 +21,17 @@ Configuration
 You can find this sample configuration file in :file:`conf/sample-node.yml`.
 
 
+Environment Variables
+~~~~~~~~~~~~~~~~~~~~~~
+
+Lymph config files support environment variable substitution for string values:
+
+.. code-block:: yaml
+
+    key: protocol://$(env.USER):$(env.PASSWORD)@host/path
+
+
+
 Dependencies
 ------------
 
--- a/lymph/config.py
+++ b/lymph/config.py
@@ -5,7 +5,7 @@ import copy
 import six
 import yaml
 
-from lymph.utils import import_object, Undefined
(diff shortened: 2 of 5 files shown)


lymph/config.py
```diff
@@ -112,8 +113,9 @@ def __repr__(self):
 
 
 class Configuration(ConfigObject):
-    def __init__(self, values=None):
+    def __init__(self, values=None, env=os.environ):
```
eguven: mutable data type as kwarg default?

emulbreh: os.environ is a global singleton, we explicitly want it to be the same object for every call.

docs/configuration.rst
```diff
@@ -21,6 +21,17 @@ Configuration
...
 
 
+Environment Variables
+~~~~~~~~~~~~~~~~~~~~~~
+
+Lymph config files support environment variable substitution for string values:
+
+.. code-block:: yaml
+
+    key: protocol://$(env.USER):$(env.PASSWORD)@host/path
```
mouadino: why env.<> ? why not use python string substitution ?

     key: protocol://{USER}:{PASSWORD}@host/path

And then we just do before parsing yaml:

    config = config.format(**os.environ)

emulbreh: string substitution is too useful as config values, e.g. 
```
greeting: "Hello {name}"
```

mouadino: Then how about using ``%(USER)s`` syntax ?

emulbreh: "string substitution is too useful in config values".

The ``$()`` syntax is extensible (might be useful to unify dependencies, load config from zk, ..) and copies a feature of circus.

eguven:
+1

PR #201, opened 2015-06-08
Use autodelete instead of exclusive 

mouadino:
Small timeouts were given us a lot of negative failures.

### DIFF
--- a/lymph/events/kombu.py
+++ b/lymph/events/kombu.py
@@ -10,6 +10,7 @@ import kombu.pools
 
 from lymph.events.base import BaseEventSystem
 from lymph.core.events import Event
+from lymph.utils.logging import setup_logger
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


lymph/events/kombu.py
```diff
@@ -81,6 +82,9 @@ def __init__(self, connection, exchange_name, serializer=DEFAULT_SERIALIZER):
         self.serializer = serializer
         self.consumers_by_queue = {}
 
+    def on_start(self):
+        setup_logger('kombu')
```
emulbreh: why do we do this on_start()? can we move this to the module level?

emulbreh:
+1, except for the setup_logger

lymph/events/kombu.py
mouadino: Usually I will say it doesn't matter really, the only problem that may come to hunt us if we put it in module level is that this will run in import time and in import time lymph logger may not be configured yet ! As for me personally I try to avoid import time side effect because they are very annoying to debug :)

FWIW This how  it's done in https://github.com/deliveryhero/lymph/blob/8cdf836e23679a3504b92c426cc725999491c21d/lymph/web/interfaces.py#L59 and https://github.com/deliveryhero/lymph/blob/68854f9494b7bf8a6f1a616213823c500174e543/lymph/discovery/zookeeper.py#L36.

PR #203, opened 2015-06-12
Correctly forward trace_ids

Drahflow:
The trace_id was extracted from incoming requests only before
spawning the new greenlet to handle the request. The new greenlet
then had no trace_id set and log messages would not be annotated
with the correct trace_id.

Also made `lymph request` generate trace_ids.

### DIFF
--- a/lymph/core/rpc.py
+++ b/lymph/core/rpc.py
@@ -186,6 +186,7 @@ class ZmqRPCServer(Component):
     def dispatch_request(self, msg):
         loglevel = self._get_loglevel(msg)
         logger.log(loglevel, '%s source=%s', msg.subject, msg.source)
+        trace.set_id(msg.headers.get('trace_id'))
         start = time.time()
         self.request_counts.incr(subject=msg.subject)
         channel = ReplyChannel(msg, self)
--- a/lymph/cli/request.py
+++ b/lymph/cli/request.py
@@ -10,6 +10,7 @@ import time
 from gevent.pool import Pool
 
 from lymph.client import Client
+from lymph.core import trace
 from lymph.exceptions import Timeout
 from lymph.cli.base import Command, handle_request_errors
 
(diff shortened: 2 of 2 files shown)


lymph/core/rpc.py
```diff
@@ -186,6 +186,7 @@ def send_reply(self, msg, body, msg_type=Message.REP, headers=None):
     def dispatch_request(self, msg):
         loglevel = self._get_loglevel(msg)
         logger.log(loglevel, '%s source=%s', msg.subject, msg.source)
+        trace.set_id(msg.headers.get('trace_id'))
```
emulbreh: Can you explain why https://github.com/deliveryhero/lymph/blob/0244277c55ad330e59540294507e1e8f99bbdef6/lymph/core/rpc.py#L202 doesn't take care of it?

Drahflow: A new greenlet is spawned in https://github.com/deliveryhero/lymph/blob/0244277c55ad330e59540294507e1e8f99bbdef6/lymph/core/rpc.py#L207. The thread_id is saved in the greenlet object. The greenlet spawned to handle the request does not have a trace_id associated.

Just apply the `lymph request` change (only), do a request across multiple services and you'll find that the trace_id is not currently forwarded.

Slightly off-topic: One could argue that *only* the newly spawned greenlet should get the trace_id (as it is this new greenlet which is processing the request). But if anything goes wrong before the greenlet is spawned, it's better to have a trace_id also set in the main greenlet, I think.

emulbreh: This should fix the propagation issue: https://github.com/deliveryhero/lymph/pull/209

emulbreh:
I like the idea of trace ids from cli command, but we should either print them somewhere or allow them to be set via arguments (e.g. `lymph request --trace=foo ...`).

emulbreh:
Follow-up: https://github.com/deliveryhero/lymph/pull/210/files

PR #204, opened 2015-06-12
Enabled environment config via yaml files

emulbreh:


lymph/config.py
```diff
@@ -111,10 +112,36 @@ def __repr__(self):
         return '%s(%r, %r)' % (self.__class__.__name__, self.root, self.path)
 
 
+_dollar_var_re = re.compile(r'\$\((\w+)\.([\w.]+)\)')
```
manisenkov: You can use named groups here, this can be more readable
`_dollar_var_re = re.compile(r'\$\((?P<first>\w+)\.(?P<second>[\w.]+)\)')`
and then below
`namespaces[match.group('first')][match.group('second')]
but it's just for interest, regexps have never been readable :)

PR #205, opened 2015-06-17
Set timeout to 3 seconds

mouadino:
Now that we have more data about response time of some rpc function
that we use internally, I believe 3 seconds is a good default


emulbreh:
definitely better than 30.
+1

PR #207, opened 2015-06-19
Make it possible to call setup_rpc_mocks more than once in setUp

mouadino:
This is usefull in inheritance since we don't want to override previous
calls of setup_rpc_mocks() from parents.


rawadtraboulsi:
:+1:  :rabbit:

PR #208, opened 2015-06-22
apply empty config for interface in instance command

mamachanko:
Call `apply_config` specified interface on in `lymph-instance` command, e.g.:

```
lymph instance --interface=examples.echo:Echo
```

### DIFF
--- a/lymph/cli/service.py
+++ b/lymph/cli/service.py
@@ -80,7 +80,8 @@ class InstanceCommand(Command):
 
         for cls_name in self.args.get('--interface', ()):
             cls = import_object(cls_name)
-            self.container.install_interface(cls)
+            instance = self.container.install_interface(cls)
+            instance.apply_config({})
 
     def _start_backdoor_terminal(self):
         # XXX(Mouad): Imported here since this is still broken in Python3.x


lymph/cli/service.py
```diff
@@ -80,7 +80,8 @@ def _setup_container(self, debug):
 
         for cls_name in self.args.get('--interface', ()):
             cls = import_object(cls_name)
-            self.container.install_interface(cls)
+            instance = self.container.install_interface(cls)
+            instance.apply_config({})
```
mouadino: This will not always work ! if interface expect a config now with this change it will in apply_config .

IMHO if you want to go in this road you can investigate whether we can pass config from command line argument, which I will suspect will need some changes in how we handle config.

emulbreh:
so someone is actually using ``--interface`` :)

mouadino:
@emulbreh I was using it with lymph-chaos, but it's very limited :(

manisenkov:
it has `--interface` argument?! O_o :smile:

emulbreh:
it was only meant for quick tests, I'd rather drop it than add more code for cli config

lymph/cli/service.py
emulbreh: We should pass an empty config view here instead of a dict.

PR #209, opened 2015-06-24
Fixed gevent pool greenlet_class settings

emulbreh:
https://github.com/gevent/gevent/blob/master/gevent/pool.py#L28


manisenkov:
:+1:

PR #210, opened 2015-06-25
Allow trace_id to be set from lymph emit/request cli

emulbreh:


mamachanko:
lgtm

mouadino:
use case ?

emulbreh:
"where did my request go?" https://github.com/deliveryhero/lymph/pull/203

mouadino:
it doesn't feel right, but maybe b/c I never need this, but I will not stand against it.

PR #211, opened 2015-06-25
Included trace_id in every http response as X-Trace-Id header

emulbreh:


mouadino:
Following on this https://github.com/deliveryhero/lymph/pull/210 wouldn't you want for the same reason to be able to send X-Trace-Id too in the request !?

emulbreh:
@mouadino yes, but incoming headers are untrusted, so we wouldn't update ``trace_id`` but rather a ``client_id``

mouadino:
I was thinking that maybe it's better to just for rpc/event request to just return trace-id and show it to user ? so no need to supply your own trace id and it should fix the problem of "where did my request go"

e.g.

    $ lymph request 'echo.echo' ... --show-trace-id
    'Hi'
    trace-id: a5ba76f...

Thoughts ?

mamachanko:
@mouadino but this is about web services, isn't it? as an external client you have no chance to identify your trace if not returned in the header.

mouadino:
@mamachanko Yes I know, and for me it seem that returning trace id is the right thing to do for both RPC and Web, instead of supplying one.

PR #213, opened 2015-06-29
Allowing installation of multiple interfaces using the same factory

dushyant88:
This allows us to install multiple interfaces using the same class but different name.

eg.

```yaml
interfaces:
    foo:
         class: interfaces:FooInterface
    bar:
         class: interfaces:FooInterace
         spam: bar
```

### DIFF
--- a/lymph/core/container.py
+++ b/lymph/core/container.py
@@ -102,7 +102,8 @@ class ServiceContainer(Componentized):
         return self.server.identity
 
     def install_interface(self, cls, **kwargs):
-        interface = self.install(cls, **kwargs)
+        interface = cls(self, **kwargs)
+        self.add_component(interface)
         self.installed_interfaces[interface.name] = interface
         for plugin in self.installed_plugins:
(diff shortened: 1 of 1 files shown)


kstrempel:
LGTM

lymph/core/components.py
```diff
@@ -92,8 +92,6 @@ def add_component(self, component):
         self.__all_components.append(component)
 
     def install(self, factory, **kwargs):
-        if factory in self._declared_components:
-            raise RuntimeError("already installed: %s" % factory)
         component = factory(self, **kwargs)
         self._declared_components[factory] = component
```
manisenkov: Does it overwrite declared component with same factory?

emulbreh:
I share @manisenkov's concern, we probably should leave the declaration stuff alone and just change https://github.com/dushyant88/lymph/blob/multiple_install/lymph/core/container.py#L105 to use add_component() and create the interface directly.

PR #215, opened 2015-07-01
Don't warn when we create sockets for random ports or without a node

emulbreh:


lymph/web/interfaces.py
```diff
@@ -47,6 +49,7 @@ def apply_config(self, config):
         self.http_port = config.get('port', self.default_http_port)
```
mouadino: Can we remove default_http_port ? since it's None.

lymph/core/container.py
```diff
@@ -114,11 +114,13 @@ def install_plugin(self, cls, **kwargs):
         return plugin
 
     def get_shared_socket_fd(self, port):
+        if 'LYMPH_SHARED_SOCKET_FDS' not in os.environ:
+            raise NoSharedSockets()
         fds = json.loads(os.environ.get('LYMPH_SHARED_SOCKET_FDS', '{}'))
```
mouadino: try:
        fds = os.environ['LYMPH_SHARED_SOCKET_FDS']
    except KeyError:
        raise NoSharedSockets()
    fds = json.loads(fds)

mouadino:
LGTM

PR #217, opened 2015-07-06
Allow config file to be specified via LYMPH_CONFIG env var

emulbreh:


lyudmildrx:
:+1: . Strange that the docs already match the change :)

eguven:
:+1:

PR #219, opened 2015-07-07
Add retry logic to lymph

mouadino:
Currently the only retrable things are NotConnected errors.

### DIFF
--- a/lymph/core/interfaces.py
+++ b/lymph/core/interfaces.py
@@ -7,6 +7,8 @@ from lymph.core.decorators import rpc, RPCBase
 from lymph.exceptions import RemoteError, EventHandlerTimeout, Timeout, Nack
 from lymph.core.components import Component, Componentized, ComponentizedBase
 from lymph.core.monitoring import metrics
+from lymph.core.retry import Retry
+from lymph.utils import make_id
 
 import gevent
 from gevent.event import AsyncResult
@@ -57,12 +59,12 @@ class ProxyMethod(object):
 
 
 class Proxy(Component):
-    def __init__(self, container, address, timeout=REQUEST_TIMEOUT, namespace='', error_map=None):
+    def __init__(self, container, address, timeout=REQUEST_TIMEOUT, retry=None, namespace='', error_map=None):
         super(Proxy, self).__init__()
         self._container = container
         self._address = address
         self._method_cache = {}
-        self._timeout = timeout
+        self._retry = retry or Retry(timeout)
         self._namespace = namespace or address
         self._error_map = error_map or {}
 
@@ -70,9 +72,9 @@ class Proxy(Component):
         self.exception_counts = metrics.TaggedCounter('exceptions', {'address': address})
 
     def _call(self, __name, **kwargs):
-        channel = self._container.send_request(self._address, __name, kwargs)
+        msg_id = make_id()
         try:
-            return channel.get(timeout=self._timeout).body
+            return self._retry.execute(self._send_request, __name, kwargs, msg_id)
         except RemoteError as e:
             error_type = str(e.__class__)
             self.exception_counts.incr(name=e.__class__.__name__)
@@ -85,6 +87,10 @@ class Proxy(Component):
         except Nack:
             self.exception_counts.incr(name='nack')
             raise
+
+    def _send_request(self, name, kwargs, msg_id):
+        channel = self._container.send_request(self._address, name, kwargs, msg_id=msg_id)
+        return channel.get().body
 
     def __getattr__(self, name):
         try:
--- a/lymph/core/retry.py
+++ b/lymph/core/retry.py
@@ -1,0 +1,31 @@
+import time
+import random
+
+import gevent
+
+from lymph.exceptions import RetryableError, Timeout
+
+
+class Retry(object):
+    retry_error = RetryableError
+    failure_error = Timeout
+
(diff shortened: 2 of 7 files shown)


lymph/core/retry.py
```diff
@@ -0,0 +1,31 @@
...
+        self._timeout = timeout
+        self._delay = delay
+        self._backoff = backoff
+        self._max_jitter = int(max_jitter * 100)
+        self._max_delay = max_delay
+
+    def execute(self, func, *args, **kwargs):
+        delay = self._delay
+        with gevent.Timeout(self._timeout, self.failure_error):
+            while 1:
```
emulbreh: True

lymph/core/retry.py
```diff
@@ -0,0 +1,31 @@
...
+
+class Retry(object):
+    retry_error = RetryableError
+    failure_error = Timeout
+
+    def __init__(self, timeout, delay=.1, backoff=2, max_jitter=.8, max_delay=None):
+        self._timeout = timeout
+        self._delay = delay
+        self._backoff = backoff
+        self._max_jitter = int(max_jitter * 100)
```
emulbreh: ``self._max_jitter = max_jitter``

lymph/core/retry.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def execute(self, func, *args, **kwargs):
+        delay = self._delay
+        with gevent.Timeout(self._timeout, self.failure_error):
+            while 1:
+                try:
+                    return func(*args, **kwargs)
+                except self.retry_error:
+                    delay *= self._backoff
+                    delay += (random.randint(0, self._max_jitter) / 100.0)
```
emulbreh: ``delay += random.random() * self._max_jitter``

emulbreh:
We shouldn't retry forever. There should be a time or count limit.

mouadino:
that;s the reason for timeout argument and ``with gevent.Timeout(...)``

emulbreh:
Ah, I didn't see that. I should go home.

lymph/core/retry.py
mouadino: For fun https://wiki.python.org/moin/WhileLoop specially this part:

>  Another version you may see of this type of loop uses 1 instead of True - this is exactly the 
> same, though in Python 2.*, using while 1 is minutely faster, due to the ability to reassign True
> to a different value - the interpretor needs to look up the value of the variable, as opposed
> to loading a constant.

lymph/core/interfaces.py
```diff
@@ -1,12 +1,15 @@
 import textwrap
 import functools
+import time
 
 import six
 
 from lymph.core.decorators import rpc, RPCBase
-from lymph.exceptions import RemoteError, EventHandlerTimeout, Timeout, Nack
+from lymph.exceptions import RemoteError, EventHandlerTimeout, Timeout, Nack, RetryableError
```
mamachanko: unused import?

lymph/core/rpc.py
```diff
@@ -151,25 +151,24 @@ def _pick_endpoint(self, service):
             if connection.is_alive():
                 choices.append(instance.endpoint)
         if not choices:
-            raise NotConnected('Not connected to %s' % service.name)
+            return
         return random.choice(choices)
 
-    def send_request(self, service, subject, body, headers=None):
+    def send_request(self, service, subject, body, headers=None, msg_id=None):
+        dst_endpoint = self._pick_endpoint(service)
```
mamachanko: `dst_endpoint` means d_estination endpoint_?

mouadino: yes, if it was not clear then I should rename, should I ?

mamachanko: `target_endpoint`, `destination`, `endpoint`, `destination_endpoint`, ...

meh. none is really good. `destination`?

mamachanko:
lgtm

lymph/core/rpc.py
```diff
@@ -151,25 +151,24 @@ def _pick_endpoint(self, service):
             if connection.is_alive():
                 choices.append(instance.endpoint)
         if not choices:
-            raise NotConnected('Not connected to %s' % service.name)
+            return
```
emulbreh: Let's keep the exception here.

lymph/core/retry.py
```diff
@@ -0,0 +1,31 @@
...
+    failure_error = Timeout
+
+    def __init__(self, timeout, delay=.1, backoff=2, max_jitter=.8, max_delay=None):
+        self._timeout = timeout
+        self._delay = delay
+        self._backoff = backoff
+        self._max_jitter = max_jitter
+        self._max_delay = max_delay
+
+    def execute(self, func, *args, **kwargs):
```
emulbreh: This could be ``__call__``

lymph/core/rpc.py
mouadino: The exception was moved to send_request, https://github.com/deliveryhero/lymph/pull/219/files#diff-e2777295f1a041ca7809f38bc2ae4547R161 and it was moved for good reason check this https://github.com/deliveryhero/lymph/pull/219/files#diff-e2777295f1a041ca7809f38bc2ae4547L170.

lymph/core/retry.py
mouadino: it could, but I like full names that express meaning, so unless you don't feel too strict about this I would like to keep execute.

mouadino:
I give up

PR #220, opened 2015-07-07
Add a simple decorator for profiling

mouadino:
@profile()
    @rpc()
    def echo(self, ...):
        ...

### DIFF
--- a/lymph/utils/profiling.py
+++ b/lymph/utils/profiling.py
@@ -1,0 +1,17 @@
+import cProfile
+import functools
+
+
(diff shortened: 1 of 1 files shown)


emulbreh:
Not sure whether this should be in lymph. Isn't there a library somewhere that has this decorator?

mouadino:
@emulbreh Didn't search for it actually, but IMHO having it in lymph make profiling any service written in lymph easily instead of having to install a library then do your profiling.

mouadino:
Not relevant anymore

PR #222, opened 2015-07-10
Only allow timezone with name

mouadino:
In case we need to serialize datetime object timezone aware with FixedOffset only
and no timezone name, now this will fail gracefully in sender side and not in receiver
side.


emulbreh:
Superseded by https://github.com/deliveryhero/lymph/pull/249/files

PR #226, opened 2015-07-14
Add support for rabbitmq cluster failover

mouadino:
- Declare queue with HA mirroring, this way rabbitmq will **do it's best** to not lose
messages by replicating them to all nodes in the cluster.
- Redeclare queues when failing over to a new node in the cluster, this will
make sure that queue and binding survive node crash.
- Add retry logic when publishing an events, this is only retry in the connection
level.
- Don't declare queues as durable, since:
   1. It doesn't mean what most things it's means.
   2. When a queue is durable, failover fail of queues.
- Don't block when getting a connection from the pool, blocking is bad at best blocking
should be used with timeout, even better fail fast.

### DIFF
--- a/lymph/events/kombu.py
+++ b/lymph/events/kombu.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 
 import sys
 from contextlib import contextmanager
+from collections import namedtuple
 
 import logging
 import kombu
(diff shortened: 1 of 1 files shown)


mouadino:
To use this PR we need to remove all existing queues, to do that easily just
reset your RabbitMQ like this:

    $ rabbitmqctl stop_app
    $ rabbitmqctl reset
    $ rabbitmqctl start_app

emulbreh:
+1

mouadino:
To delete all queue we can also do:

    for queue in $(rabbitmqadmin list queues name -f bash); do rabbitmqadmin delete queue name=$queue; done

PR #233, opened 2015-07-24
set process title after registering

eguven:
Before:

    lymph-instance (identity: None, endpoint: None, config: ./config.yml)

After:

    lymph-instance (identity: 9e757f2118ae87d4702b6b483263d199, endpoint: tcp://10.13.220.99:64669, config: ./config.yml)


mouadino:
LGTM

PR #236, opened 2015-08-07
Don't fail when machine go to sleep

mouadino:
If dev machine go to sleep (lymph instance and zookeeper in same machine) as
soon as machine is back lymph will try to re-register instance with Zookeeper
which may fail with NodeExistsError if zookeeper also went to sleep and this later
didn't get the chance to detect broken connection and delete ephemeral node.


mamachanko:
lgtm

PR #237, opened 2015-08-07
Better process metrics

mouadino:
Use psutil to generate process global metrics e.g. cpu/memory/files/io ... .

The problem with rusage is that it's too low level, hard to understand and also
sometimes hard to use since it depend on the platform example:

maxrss which is in kilobytes in Lunix and bytes in OSX.
    - http://man7.org/linux/man-pages/man2/getrusage.2.html
    - https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man2/getrusage.2.html

This new process metric aim to give more fine grain and understandable metrics.


emulbreh:
+1

PR #240, opened 2015-08-07
Accept request body from stdin

mouadino:
When body is too big usually you will have it in a file a pretty version
that you will edit and in that case it way easier to do:

    cat body.json | lymph request service.func -

Than the alternative which is copy-paste.

### DIFF
--- a/lymph/cli/request.py
+++ b/lymph/cli/request.py
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 
 class RequestCommand(Command):
     """
-    Usage: lymph request [options] <subject> <params>
+    Usage: lymph request [options] <subject> <params> [-]
 
     Description:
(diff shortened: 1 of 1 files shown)


stefano-dh:
:+1: :tropical_drink:

mamachanko:
Including the example in the command's description would be very helpful. Not everyone's familiar with the `-` convention.

mamachanko:
https://github.com/deliveryhero/lymph/pull/242

lymph/cli/request.py
```diff
@@ -20,7 +20,7 @@
 
 class RequestCommand(Command):
     """
-    Usage: lymph request [options] <subject> <params>
+    Usage: lymph request [options] <subject> <params> [-]
```
emulbreh: this will accept ``params`` and dash

PR #241, opened 2015-08-07
Add useful rpc functions to the node interface

mouadino:
It's useful to get information about processes running by calling the node
rpc interface. Example when it can be useful is lymph-chaos (
https://github.com/mouadino/lymph-chaos)

PR #245, opened 2015-08-20
Publish monitoring data on random port

emulbreh:


lymph/services/node.py
```diff
@@ -61,8 +61,6 @@ def _get_metrics(self):
 
 
 class Node(Interface):
-    register_with_coordinator = False
```
mouadino: why registering node ?

lymph/core/container.py
```diff
@@ -68,7 +67,7 @@ def __init__(self, rpc=None, registry=None, events=None, log_endpoint=None, serv
             self.add_component(self.events)
             self.events.install(self)
 
-        self.monitor = self.install(MonitorPusher, aggregator=self.metrics_aggregator, endpoint=self.monitor_endpoint, interval=5)
+        self.monitor = self.install(MonitorPusher, aggregator=self.metrics_aggregator, endpoint=rpc.ip, interval=5)
```
mouadino: Probably not needed for now, but may be good to also make monitoring ip different than RPC for multiple NICs setup.

lymph/services/node.py
emulbreh: Because there's no reason not to. I removed it for the default interface and Node was the only remaining unregistered interface.

mouadino: which name the node will it be registered under ?

emulbreh: `node`

emulbreh: It's also useful for https://github.com/deliveryhero/lymph/pull/241

mouadino:
LGTM

PR #246, opened 2015-08-24
Set gevent version

vpikulik:


setup.py
```diff
@@ -11,7 +11,7 @@
     'docopt>=0.6.1',
     'kazoo>=1.3.1',
     'kombu>=3.0.16',
-    'gevent',
+    'gevent==1.1b3',
```
mouadino: usually we don't pin here we use >= as for why ask @emulbreh

mouadino:
I don't like using beta quality dependencies b/c it make all lymph beta quality too :(

setup.py
emulbreh: yes, no pinning in setup.py.
https://caremad.io/2013/07/setup-vs-requirement/

vpikulik: I removed this change

PR #247, opened 2015-08-24
Added hostname to interface description

emulbreh:


jacqueslh:
LGTM

lymph/core/container.py
```diff
@@ -145,6 +145,7 @@ def get_instance_description(self, interface):
             'log_endpoint': self.log_endpoint,
             'backdoor_endpoint': self.backdoor_endpoint,
             'fqdn': self.fqdn,
+            'hostname': socket.gethostname(),
```
mouadino: why not ?

    'hostname': self.fqdn

mouadino: fqdn is already there why we want hostname ?

emulbreh: Eren wants the hostname (since fqdn is somewhat randomly defined as the first name that contains a dot) https://docs.python.org/2/library/socket.html#socket.getfqdn

PR #248, opened 2015-08-25
Enable configuring WSGI middlewares with lymph web

mouadino:


docs/api/web_api.rst
```diff
@@ -7,13 +7,49 @@ Web API
...
+              class: module:CORSMiddleware
+              origin: http://deliveryhero.com
+              methods:
+                 - GET
+
+        interfaces:
+            web:
+                class: web:JsonrpcGateway
+                middlewares:
+                  - cors
```
emulbreh: Let's make this `dep:cors` so you can declare middleware inline.

lymph/web/middleware.py
```diff
@@ -0,0 +1,18 @@
+import abc
+
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class WSGIMiddleware(object):
```
emulbreh: Let's use the reglar wsgi middleware pattern instead. That allows 3rd party middleware to be used without wrapping:
```
decorated_app = middleware(app)
```

docs/api/web_api.rst
```diff
@@ -7,13 +7,49 @@ Web API
...
 
 
-.. _werkzeug.routing.Map: http://werkzeug.pocoo.org/docs/0.10/routing/#maps-rules-and-adapters
\ No newline at end of file
+Web Configuration
+==================
+
+.. describe:: port
+
+    Static port to listen to. If not supplied lymph will bind to a port dynamically.
```
mamachanko: "to a _random_ port" maybe

docs/api/web_api.rst
```diff
@@ -7,13 +7,49 @@ Web API
...
+
+.. describe:: wsgi_pool_size
+
+    Gevent HTTP server Greenlets pool size.
+    Default: None, which mean no limit.
```
mamachanko: means

docs/api/web_api.rst
```diff
@@ -7,13 +7,49 @@ Web API
...
+
+.. describe:: middlewares
+
+    List of WSGI middlewares to use. For a WSGI middleware to be compatible with
+    lymph web it must implement the ``lymph.web.middleware:WSGIMiddleware``
```
mamachanko: ~~the~~

lymph/web/middleware.py
mouadino: And how are you going to configure middlewares ?

emulbreh: In the constructor:
```
app = MiddlewareClass(config)(app)
```

mouadino: MiddlewareClass ``__call__`` should implement the wsgi protocol.

lymph/config.py
```diff
@@ -69,15 +66,25 @@ def _create_instance(self, key, instance_config, default_class=None, **kwargs):
             return cls(**instance_config)
 
     def get_instance(self, key, default_class=None, **kwargs):
-        instance_data = self.get(key)
-        if isinstance(instance_data, six.string_types) and instance_data.startswith('dep:'):
-            _, dep_name = instance_data.split(':', 1)
+        instance_config = self.get(key)
+        return self._get_instance(key, instance_config, default_class=default_class, **kwargs)
+
+    def get_instance_collection(self, key, default_class=None, **kwargs):
```
mamachanko: Please, include in the [config API docs](https://raw.githubusercontent.com/deliveryhero/lymph/master/docs/api/config_api.rst).

lymph/web/middleware.py
mouadino: Just to be more clear, I already thought about this and b/c most existing community middleware are in the form:

    class MyMiddleware(object):
           def __init__(self, app, *middleware_options):
                 ...

It's hard to reuse them without adapting them e.g. http://werkzeug.pocoo.org/docs/0.10/middlewares/.

lymph/config.py
mouadino: Done

lymph/web/middleware.py
mamachanko: It could only think of something along the lines of `apply_config` to fulfill both requirements.

mouadino: @mamachanko I don't see how apply_config is a solution to use 3rd party middleware without wrapping ?

mamachanko: @mouadino yes, bare usage wouldn't work then

lymph/config.py
```diff
@@ -47,9 +47,6 @@ def create_instance(self, key, default_class=None, **kwargs):
         return self._create_instance(key, instance_config, default_class=default_class, **kwargs)
 
     def _create_instance(self, key, instance_config, default_class=None, **kwargs):
-        if instance_config is None:
-            raise ConfigurationError("no config available for %r" % key)
-
```
emulbreh: why remove the exception?

lymph/config.py
```diff
@@ -69,15 +66,26 @@ def _create_instance(self, key, instance_config, default_class=None, **kwargs):
             return cls(**instance_config)
 
     def get_instance(self, key, default_class=None, **kwargs):
-        instance_data = self.get(key)
-        if isinstance(instance_data, six.string_types) and instance_data.startswith('dep:'):
-            _, dep_name = instance_data.split(':', 1)
+        instance_config = self.get(key)
+        return self._get_instance(key, instance_config, default_class=default_class, **kwargs)
+
+    def get_instances_collection(self, key, default_class=None, **kwargs):
```
emulbreh: can we name this just ``get_instances()``?

lymph/web/middleware.py
emulbreh: Even if we need something like a middleware factory for configuration I'd prefer something that just uses ``__call__``:
```
class MiddlewareFactor(object):
    def __init__(**config):
        ...
    def __call__(self, app):
        return wrapped_app
```

emulbreh:
+1, except for the removed exception

lymph/web/middleware.py
mouadino: That's not a WSGI Middleware :)

A WSGI Middleware must be a WSGI application it self. Please check https://www.python.org/dev/peps/pep-3333/#middleware-components-that-play-both-sides.

Else if you are arguing that ``decorate`` should be renamed to ``__call__`` and introduce a new indirection where ``wrapped_app`` is the real middleware then I don't see what all this indirection is suppose to solve ?

BTW what you are proposing still doesn't solve the "allows 3rd party middleware to be used without wrapping".

lymph/config.py
mouadino: b/c I am stupid :)

lymph/config.py
mouadino: The problem with two different functions with only one char different is that they are hard to tell the difference between them i.e. ``get_instance()`` vs ``get_instances()``. 

For ``get_instance_list`` shouldn't be ``get_instances_list`` ? plural ?

docs/api/web_api.rst
mouadino: I revert to not have inlined middlewares b/c they are ugly and it just make dealing with configuration much easier.

By ugly I mean:

      interface:
           web:
                 middlewares:
                        - class: module:CORSMiddleware
                           origin: http://deliveryhero.com
                            methods:
                                  - GET

No one should ever have to do this :)

emulbreh: Let's only have one way to access dependencies, please. You don't need to declare middleware inline, but there's no reason to force the use of dependencies here.

mouadino: Please to you back, this PR is standing here for 3 months, please let's merge it, if you fell strong about it, you can do it, it's an extra feature, if can be done later without breaking compatibility.

emulbreh: It's been three months because we didn't work on it after agreeing on the approach. There's no urgency to get it merged.

PR #249, opened 2015-08-27
Making the DateTime serializer work of iso8601

dushyant88:


vpikulik:
:+1:

mei-li:
Can we bump the version after that?

PR #250, opened 2015-08-31
Use different gevent versions per python version

kpanic:


setup.py
```diff
@@ -11,7 +11,6 @@
     'docopt>=0.6.1',
     'kazoo>=1.3.1',
     'kombu>=3.0.16',
-    'gevent',
```
mouadino: we don't pin in setup.py, the previous version should work just fine.

PR #254, opened 2015-09-07
Not found hook for web interfaces

lyudmildrx:
Update web interfaces. Add a not found hook that interfaces can overwrite. This can be useful when http interfaces that want to stick to unified response content type different than html

### DIFF
--- a/lymph/tests/integration/test_web_interface.py
+++ b/lymph/tests/integration/test_web_interface.py
@@ -1,3 +1,4 @@
+from werkzeug.exceptions import NotFound, MethodNotAllowed
 from werkzeug.routing import Map, Rule
 from werkzeug.wrappers import Response
 
@@ -17,10 +18,22 @@ class HandledRuleHandler(RequestHandler):
         return Response("Handled Rule Handler")
 
 
+class CustomNotFound(NotFound):
+    def __init__(self):
+        response = Response("never-gonna-match-you-down", status=404)
+        super(CustomNotFound, self).__init__(response=response)
+
+
+class CustomMethodNotAllowed(MethodNotAllowed):
+    def get_body(self, *args, **kwargs):
+        return "never-gonna-run-around-or-post-you"
+
+
 class Web(WebServiceInterface):
     url_map = Map([
         Rule("/test/", endpoint="test"),
-        Rule("/foo/", endpoint=RuleHandler),
+        Rule("/foo/", endpoint=RuleHandler, methods=['get']),
+        Rule("/baz/", endpoint=RuleHandler),
         HandledRule("/bar/", endpoint="bar", handler=HandledRuleHandler),
         Rule("/fail/", endpoint="fail"),
         Rule("/fail-wrong-endpoint/", endpoint=42),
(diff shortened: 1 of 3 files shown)


mouadino:
when is user suppose to override this "hook" ? is it for all NotFound(s) ? or when code raise NotFound ?

lyudmildrx:
Domain is hit but path is not recognised. If path is recognised handler is called and is responsible to raise 404 if there is a resource not found

mouadino:
The the try-except should only be around https://github.com/deliveryhero/lymph/pull/254/files#diff-ca2df89c6a5b4dbf735e918158ae6cccR87 ?

mouadino:
you may also want to handle other exceptions like MethodNotAllowed http://werkzeug.pocoo.org/docs/0.10/routing/#werkzeug.routing.MapAdapter.match ?

lymph/web/interfaces.py
```diff
@@ -85,36 +85,44 @@ def dispatch_request(self, request):
...
-                if self.container.debug:
-                    raise
-                response = Response('', status=500)
-        response.headers['X-Trace-Id'] = trace.get_id()
+        except NotFound as e:
+            response = self.not_found(request, e)
+        except MethodNotAllowed as e:
+            response = self.not_allowed(request, e)
+        else:
+            try:
```
mouadino: can you move this into it's own method ? def handle(self, request, rule, kwargs):  ?

mouadino:
I don't really like override method as an API, since user in the end will repeat same boilderplate.

    def not_found(self, request, exception):
          # I have now to ignore exception.
          return MyNotFound().from_response()

How about something like this ?

     class WebInterface(...):
            NotFound = werkzeug.exceptions.NotFound
     
Then user can just supply his own NotFound when inheriting from WebInterface.

Thoughts ?

lyudmildrx:
A bit more limiting but cleaner and would kind of force usage of werkzeug like exceptions.

mouadino:
LGTM

Would prefer it better if you can squash commits before merging :)

lymph/tests/integration/test_web_interface.py
```diff
@@ -58,3 +58,7 @@ def test_dispatch_failing_endpoint_to_500(self):
         response = self.client.get("/fail-wrong-endpoint/")
         self.assertEqual(response.data.decode("utf8"), "")
         self.assertEqual(response.status_code, 500)
+
+    def test_dispatch_not_found(self):
```
mouadino: Can you add a test of overriding NotFound ?

lymph/tests/integration/test_web_interface.py
```diff
@@ -20,7 +23,8 @@ def get(self):
 class Web(WebServiceInterface):
```
mouadino: Can't we define our own NotFound and MethodNotAllowed here instead of mocking ?

lyudmildrx: Can not test default then or need to add "another Web" and another testcase with that service class.
Is mocking bad?

mouadino: My philosophy "if I can avoid mocking I do."

I don't it's too bad to add another interface, since that nothing in term of lines of code and give good example as to who it's expected to be used. Right ?

lyudmildrx:
@mouadino tests done. I figured all tests could be run with a changed error handlers. To be sure

mouadino:
LGTM

PR #255, opened 2015-09-07
Fix autodoc for raises argument.

mouadino:
- raises may be a single exceptions and not always a tuple.
- Include also docs for exceptions.


emulbreh:
+1

Menda:
:+1:

PR #256, opened 2015-09-07
Stick to werkzeug exception interface

lyudmildrx:
Do not expect exception to define a to_http_response method, but
rather stay with werkzeugs get_response. This change avoids
unnecessary complexity in the interface.


mouadino:
@lyudmildrx https://github.com/deliveryhero/lymph/pull/64 it was @emulbreh decision to have another name :)

lyudmildrx:
Closing as change would not only force all web interfaces to use wergzeug style exceptions, but exactly wergzeug exceptions

PR #259, opened 2015-10-08
Report EventHandler/TaskHandler calls as transaction in newrelic

mouadino:


mouadino:
Couldn't find a nicer way, and yes we should get rid of Declaration.

emulbreh:
+1

PR #261, opened 2015-10-13
Retryable event handling

mouadino:
Add a retry option to ``lymph.event(...)`` that specify max number of
reties to do in case an exception is raised when handling the given event.
The way retry work is by republishing the event to the handler queue while setting
a header that specify the number of reties left, this way of requeuing an event
at the opposite to ``message.reject(requeue=True)`` enable us to set a limit of the
numbers of retries.

Note: This feature should only be used when the event handler is guarantee to be idempotent.

### DIFF
--- a/lymph/events/kombu.py
+++ b/lymph/events/kombu.py
@@ -20,6 +20,7 @@ logger = logging.getLogger(__name__)
 DEFAULT_SERIALIZER = 'lymph-msgpack'
 DEFAULT_EXCHANGE = 'lymph'
 DEFAULT_MAX_RETRIES = 3
+RETRY_HEADER = 'retry_count'
 
 # Info of where the queue master server, so that we can redeclare the queue
 # when we failover to another master.
@@ -35,6 +36,11 @@ class EventConsumer(kombu.mixins.ConsumerMixin):
         self.greenlet = None
         self.event_system = event_system
         self.connect_max_retries = max_retries
+        self.retry_producer = EventProducer(
+            exchange=self.event_system.retry_exchange,
+            routing_key=handler.queue_name,
+            event_system=event_system,
+        )
 
     def get_consumers(self, Consumer, channel):
         return [Consumer(queues=[self.queue], callbacks=[self.on_kombu_message])]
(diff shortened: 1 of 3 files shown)


lymph/events/kombu.py
```diff
@@ -67,15 +74,23 @@ def _handle_message(self, body, message):
...
-            message.reject()
-            self.event_system.container.error_hook(sys.exc_info())
+            self._handle_fail(message, event)
         finally:
             if self.handler.once:
                 self.event_system.unsubscribe(self.handler)
 
+    def _handle_fail(self, message, event):
+        retry = event.headers.get(RETRY_HEADER, self.handler.retry) - 1
+        if retry > 0:
```
emulbreh: shouldn't this be `>=0`?

lymph/events/kombu.py
```diff
@@ -20,6 +20,7 @@
 DEFAULT_SERIALIZER = 'lymph-msgpack'
 DEFAULT_EXCHANGE = 'lymph'
 DEFAULT_MAX_RETRIES = 3
+RETRY_HEADER = 'x-retry-count'
```
emulbreh: Let's stick to Python identifiers for headers (we have `trace_id` already)

lymph/tests/integration/test_kombu_events.py
```diff
@@ -94,3 +99,7 @@ def test_broadcast_event(self):
         self.assert_eventually_true(self.received_broadcast_check(2), timeout=10)
         self.assertEqual(self.the_interface.collected_events[0].evt_type, 'foo_broadcast')
         self.assertEqual(self.the_interface_broadcast.collected_events[0].evt_type, 'foo_broadcast')
+
+    def test_retryable_event(self):
+        self.lymph_client.emit('retryable_foo', {})
+        self.assert_eventually_true(self.received_check(2), timeout=10)
```
emulbreh: 3 (one original event + two retries)

mouadino:
@emulbreh Comments addressed

emulbreh:
+1

PR #262, opened 2015-10-14
Make the jsonrpc.html file a bit more complete with an interactive form.

fenfir:
I added some input fields to the root page to make interacting with the jsonrpc provider easier.


emulbreh:
Thanks

PR #267, opened 2015-10-28
Deal with zookeeper connection issues more gracefully

emulbreh:


lymph/discovery/zookeeper.py
```diff
@@ -56,6 +56,10 @@ def on_kazoo_state_change(self, state):
                 self.spawn(self.register, name, instance)
             for service in six.itervalues(self.cache):
                 self.spawn(self.lookup, service, timeout=None)
+        elif state == KazooState.LOST:
+            logger.warning('zookeeper connection lost')
+        elif state == KazooState.SUSPENDED:
+            logger.info('zookeeper connection suspended')
```
mouadino: else:
          logger.info('zookeeper connection state changed to %s', state)

lymph/discovery/zookeeper.py
```diff
@@ -99,9 +105,9 @@ def lookup(self, service, timeout=1):
         try:
             names = result.get(timeout=timeout)
         except NoNodeError:
-            raise LookupFailure("failed to resolve %s" % service.name)
-        except ConnectionLoss:
-            logger.warning("lost zookeeper connection")
+            raise LookupFailure("failed to resolve %s", service.name)
+        except KazooException as e:
+            logger.info("zookeeper lookup failure: %s", e)
```
mouadino: info ? IMHO this should be at least a warning.

lymph/discovery/zookeeper.py
emulbreh: That shouldn't happen, zk can only have three states: connected, suspended, lost.

mouadino:
LGTM

PR #269, opened 2015-10-29
Added RPC calls to newrelic traces

emulbreh:


lymph/plugins/newrelic.py
```diff
@@ -25,8 +37,8 @@ def __init__(self, container, config_file=None, environment=None, **kwargs):
         self.container.error_hook.install(self.on_error)
         self.container.http_request_hook.install(self.on_http_request)
         newrelic.agent.initialize(config_file, environment)
-        for method in ('send_request', 'emit_event', 'lookup'):
-            setattr(ServiceContainer, method, newrelic.agent.function_trace()(getattr(ServiceContainer, method)))
+        RequestChannel.get = trace_rpc_method(RequestChannel.get, lambda channel: channel.request.subject)
+        DeferredReply.get = trace_rpc_method(DeferredReply.get, lambda deferred: deferred.subject)
```
mouadino: This will only show the time that you wait for async result to return correct ?

emulbreh: Yes. It would have been the same with only RequestChannel.get

mouadino: than the time that will be shown is not how much RPC request spend but how much time we waited for it, and it worst with deferred call :( Is there a good reason for showing a wrong/misleading time ?

emulbreh: as far as I understand newrelic traces are always sequential.

mouadino:
+1

PR #275, opened 2015-11-16
Added generic http healthcheck endpoint

emulbreh:


manisenkov:
:+1:

lymph/web/interfaces.py
```diff
@@ -146,3 +150,9 @@ def get_handler(self, request, rule):
 
     def get_wsgi_application(self):
         return self.application
+
+    def is_healthy(self):
```
ramovsky: @propery maybe

lymph/web/interfaces.py
```diff
@@ -146,3 +150,9 @@ def get_handler(self, request, rule):
...
+        return True
+
+    def get_health_check_response(self, request):
+        return Response(status=200 if self.is_healthy() else 503)
```
mouadino: why 503 ? probably it doesn't matter but since you went all thing length to make it configurable (for what ever reason :)), how about:

```
def healthy(self):
    return 200

def get_health_check_response(self, request):
    return Response(status=self.healthy())
```

emulbreh: I wanted to have a clear convention (defaults that are hard to screw up). 200 is obviously healthy, and from the available status codes, 503 looked most reasonable for the unhealthy case: "The server is currently unable to handle the request due to a temporary overloading or maintenance of the server."

eguven: how about `418 I'm a teapot (RFC 2324)` :smiling_imp:

emulbreh: come on, it's a problem on the server side, so 5xx instead of 4xx.

kstrempel:
:+1:

PR #278, opened 2015-11-24
Added newrelic app_name plugin config

emulbreh:


lymph/plugins/newrelic.py
```diff
@@ -31,12 +32,20 @@ def wrapped(self, *args, **kwargs):
...
 class NewrelicPlugin(Plugin):
-    def __init__(self, container, config_file=None, environment=None, **kwargs):
+    def __init__(self, container, config_file=None, environment=None, app_name=None, **kwargs):
         super(NewrelicPlugin, self).__init__()
         self.container = container
         self.container.error_hook.install(self.on_error)
         self.container.http_request_hook.install(self.on_http_request)
         newrelic.agent.initialize(config_file, environment)
+
+        settings = newrelic.agent.global_settings()
```
mouadino: shouldn't this be in newrelic.ini ?

emulbreh: we would need a newrelic.ini file for every service then.

mouadino:
LGTM

PR #279, opened 2015-11-25
Add cross application tracing using newrelic

mouadino:
- Get rid of monkey patching by making needed component extendable (observer pattern)
- Add support for newrelic cross application tracing replacing normal function tracing for rpc
- Add new tracing package that allow pluggeable tracing scenarios to allow developer to instrument
their services (for now only newrelic is supported)


lymph/core/channels.py
```diff
@@ -35,6 +37,13 @@ def get(self, timeout=1):
     def close(self):
         del self.server.channels[self.request.id]
 
+    def on_new_message(self, func):
+        self._callbacks.append(func)
+
+    def _tigger_callbacks(self, msg):
```
manisenkov: typo

emulbreh:
lymph.tracing seems to be unused and undocumented.

mouadino:
Unused that's normal since it's new and it meant to be used by service writers :)
Undocumented b/c I would like to see how it will get used first in some of our service before publishing the API, since changing a published API in lymph is very hard.

Basically yes, for now only ``trace_context`` is supplied, and what service writer have to do is pass the tracer to the component that they would like to trace in depth, and use it there.

mouadino:
Putting this in hold since I don't want to waste more time on this.

PR #280, opened 2015-12-03
RPC Versioning

emulbreh:


lymph/core/container.py
```diff
@@ -42,6 +45,32 @@ class InterfaceSkipped(Exception):
...
+            self.latest = interface
+        self.versions[interface.version] = interface
+
+    def __iter__(self):
+        return six.itervalues(self.versions)
+
+    def __str__(self):
+        return ', '.join(str(v) for v in self.versions.keys())
+
+    def get(self, version):
```
mouadino: ``def get(self, version=None)`` ?

lymph/core/container.py
```diff
@@ -136,9 +166,15 @@ def get_shared_socket_fd(self, port):
         except KeyError:
             raise SocketNotCreated()
 
+    def iter_interfaces(self):
+        for versions in six.itervalues(self.installed_interfaces):
```
mouadino: ``for interfaces in six.iter...``

lymph/config.py
```diff
@@ -103,22 +103,22 @@ def as_dict(self):
         return self.root.get_raw(self.path)
 
     def get_raw(self, key, default=Undefined):
-        return self.root.get_raw('%s.%s' % (self.path, key), default)
+        return self.root.get_raw(self.path + [key], default)
```
mouadino: That's why I don't like . as separators :)

lymph/core/channels.py
```diff
@@ -40,25 +40,29 @@ class ReplyChannel(Channel):
     def __init__(self, request, server):
         super(ReplyChannel, self).__init__(request, server)
         self._sent_reply = False
+        self._headers = {}
+
+    def add_header(self, name, value):
```
mouadino: This looks weird, what happen if I add_header after reply is send ?

lymph/core/container.py
```diff
@@ -136,9 +166,15 @@ def get_shared_socket_fd(self, port):
...
+    def iter_interfaces(self):
+        for interfaces in six.itervalues(self.installed_interfaces):
+            for interface in interfaces:
+                yield interface
+
     @property
     def service_types(self):
-        return self.installed_interfaces.keys()
+        for interface in self.iter_interfaces():
+            yield '%s@%s' % (interface.name, interface.version)
```
mouadino: ``yield interface.versionned_name`` ?

The '@' look likes a magic constant here.

lymph/core/channels.py
emulbreh: headers are obviously only included for all messages that you send afterwards. I'm not happy with it either but couldn't find a better place to add it because the rpc implementation doesn't know which interface ends up using the channel. That should probably be refactored, but not in this PR.

lymph/core/container.py
emulbreh: No need, it's always called with a version.

lymph/core/container.py
```diff
@@ -205,16 +247,25 @@ def emit_event(self, event_type, payload, headers=None, **kwargs):
...
+            version = semantic_version.Version(version)
+        try:
+            versions = self.installed_interfaces[interface_name]
+        except KeyError:
+            logger.warning('Unsupported interface: %s', interface_name)
+            channel.nack(True)
+            return
         try:
-            interface = self.installed_interfaces[interface_name]
+            interface = versions.get(version)
```
mouadino: shouldn't this be ``versions[version]`` ?

lymph/core/container.py
emulbreh: The "@" is used in a couple of places, and not all of them have an interface. I don't see the advantage of using a constant here. It's mostly punctuation, except for the parsing part in lymph.core.versioning.

lymph/core/rpc.py
```diff
@@ -136,44 +131,53 @@ def _send_message(self, endpoint, msg):
         logger.debug('-> %s to %s', msg, endpoint)
         connection.on_send(msg)
 
-    def prepare_headers(self, headers):
-        headers = headers or {}
+    def prepare_headers(self, headers, **extra_headers):
```
mouadino: Why not just ``prepare_headers(headers)`` ?

lymph/core/container.py
emulbreh: versions is an instance of InterfaceVersions, which doesn't implement `__getitem__`. Maybe it's a bit more obvious, I'll change it.

lymph/core/rpc.py
```diff
@@ -136,44 +131,53 @@ def _send_message(self, endpoint, msg):
...
             if connection.is_alive():
-                choices.append(instance.endpoint)
+                choices.append(instance)
         if not choices:
-            raise NotConnected('Not connected to %s' % service.name)
+            raise NotConnected('Not connected to %s' % service)
         return random.choice(choices)
 
     def send_request(self, service, subject, body, headers=None):
+        if isinstance(service, InstanceSet):
```
mouadino: can we get rid of this ? by making service always ``InstanceSet`` or mocking _pick_instance to ``Service[Set]`` ?

emulbreh: that would be nice, but I'd rather not do it in this pr. `service` can be a string here.

lymph/core/rpc.py
emulbreh: headers can be `None`. Of course we could also move out this check, but I'd rather have that factored out.
```
headers = headers or {}
headers['version'] = foo
self.prepare_headers(headers)
```
vs
```
self.prepare_headers(headers, version=foo)
```

lymph/core/versioning.py
```diff
@@ -0,0 +1,24 @@
+from semantic_version import Version, Spec
+import six
+
+
+def parse_versioned_name(name):
```
mouadino: can we have some unit test for this ?

lymph/discovery/zookeeper.py
```diff
@@ -101,31 +101,30 @@ def lookup(self, service, timeout=1):
...
-            identity = kwargs.pop('identity')
-            service.update(identity, **kwargs)
+        logger.info("lookup %s %r", service_name, instance_ids)
+        old_instance_ids = {instance.id for instance in service}
+        for instance_id in instance_ids:
+            kwargs = self._get_service_znode(service, service_name, instance_id)
+            service.update(instance_id, **kwargs)
             try:
-                identities.remove(identity)
+                old_instance_ids.pop()
```
mouadino: I don't understand this!

lymph/tests/test_rpc_versioning.py
```diff
@@ -0,0 +1,63 @@
...
+    containers = [
+        {
+            'foo@1.1': {'class': Foo11},
+            'foo@1.5': {'class': Foo15},
+            'foo@2.1': {'class': Foo21},
+        },
+        {
+            'foo@1.2': {'class': Foo12},
+            'foo@3.3': {'class': Foo33},
+            'foo@4.0': {'class': Foo33},
```
mouadino: s/Foo33/Foo40

lymph/discovery/zookeeper.py
emulbreh: it does the same it did before: go through discovered ids, update service object, remove instances that are no longer there.

lymph/tests/test_rpc_versioning.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def test_version_matching(self):
+        proxy = self.client.proxy('foo', version='1.1')
+        self.assertIn(proxy.get_class_name(), {'Foo11', 'Foo12', 'Foo15'})
+
+        proxy = self.client.proxy('foo', version='1.2')
+        self.assertIn(proxy.get_class_name(), {'Foo12', 'Foo15'})
+
+        proxy = self.client.proxy('foo', version='1.7')
+        self.assertRaises(Exception, proxy.get_class_name)
```
mouadino: Exception looks too generic, can we raise something better or check exception message?

lymph/tests/test_rpc_versioning.py
emulbreh: intentional, it should be possible to expose the same interface class for multiple versions

mouadino:
How does this looks when using ``lymph discover``.

emulbreh:
@mouadino discover looks the same as before, I have a branch on top with cli fixes: https://github.com/emulbreh/lymph/compare/versioning...emulbreh:cli-improvements?expand=1

lymph/core/versioning.py
emulbreh: It's 100% covered.

lymph/core/container.py
mouadino: It's not about the constant it's about the fact that the versioned name of the interface should belong to interface, not constructed outside it, e.g. imagine if we want to change the syntax of versioning e.g. echo:1.2.3 :)

lymph/discovery/zookeeper.py
mouadino: but old_instance_ids is a set right ? and old_instance_ids.pop() remove a random element for the set, or I am missing something ?

lymph/core/rpc.py
mouadino: I will prefer version 1, btw ``headers = headers or {}`` is done anyway it's just a matter of doing it before calling the method.

lymph/tests/test_rpc_versioning.py
mouadino: b/c this is confusing https://github.com/deliveryhero/lymph/pull/280/files#diff-4002733f35ff273740c391e04a4899efR64

lymph/core/rpc.py
emulbreh: ok

mouadino:
LGTM

mouadino:
@emulbreh Thanks for making the specification reality :)

Issue #287, opened 2016-01-04
WebInterface should check for correct return type


eguven:
lymph fails when setting trace-id header in case the return type from WebInterface is not correct type. lymph should probably check for this and raise before.

    Traceback (most recent call last):
      File "/home/eren.gueven/.virtualenvs/ars/local/lib/python2.7/site-packages/gevent/pywsgi.py", line 508, in handle_one_response
        self.run_application()
      File "/home/eren.gueven/.virtualenvs/ars/local/lib/python2.7/site-packages/gevent/pywsgi.py", line 494, in run_application
        self.result = self.application(self.environ, self.start_response)
      File "/home/eren.gueven/.virtualenvs/ars/local/lib/python2.7/site-packages/werkzeug/wrappers.py", line 291, in application
        return f(*args[:-2] + (request,))(*args[-2:])
      File "/home/eren.gueven/.virtualenvs/ars/local/lib/python2.7/site-packages/lymph/web/interfaces.py", line 114, in dispatch_request
        response.headers['X-Trace-Id'] = trace.get_id()
    AttributeError: 'NoneType' object has no attribute 'headers'

emulbreh:
Fixed in #288

PR #288, opened 2016-01-04
Fail cleanly when request handlers don't return response objects

emulbreh:
https://github.com/deliveryhero/lymph/issues/287

### DIFF
--- a/lymph/web/interfaces.py
+++ b/lymph/web/interfaces.py
@@ -122,6 +122,8 @@ class WebServiceInterface(Interface):
                 response = handler.dispatch(kwargs)
             else:
                 response = handler(request, **kwargs)
+            if not isinstance(response, Response):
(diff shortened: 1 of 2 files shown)


lymph/web/interfaces.py
```diff
@@ -122,6 +122,8 @@ def handle(self, request, rule, kwargs):
                 response = handler.dispatch(kwargs)
             else:
                 response = handler(request, **kwargs)
+            if not isinstance(response, Response):
```
mouadino: not fan of this kind of checks, IMHO the error raises in the issue #287 is already good enough, checking return type like this is not pythonic, it's the same argument of whether you would check parameters.

emulbreh: This check is necessary to make sure you get a trace id in this case.

eguven:
:+1:

PR #291, opened 2016-01-11
Extensible tracing

emulbreh:


mamachanko:
As far as I can tell it looks good. Docs?

docs/api/trace_api.rst
```diff
@@ -0,0 +1,16 @@
+.. currentmodule:: lymph.core.trace
+
+
+Trace API
+=========
+
+.. function:: context(trace_id=None)
```
mouadino: can we specify that we use gevent thread local for persisting trace id ? b/c a context manager that return nothing is always magical (done for the side effect).

docs/api/trace_api.rst
```diff
@@ -0,0 +1,16 @@
...
+
+    Returns a context manager that will set the trace id. 
+    If ``trace_id`` is None, a new random id will be generated.
+    
+
+.. function:: get_id()
+
+    Returns the current trace id.
+
```
mouadino: .. function: get_headers() ?

mouadino: from_headers ?

lymph/plugins/newrelic.py
```diff
@@ -32,14 +32,16 @@ def wrapped(self, *args, **kwargs):
...
 class NewrelicPlugin(Plugin):
-    def __init__(self, container, config_file=None, environment=None, app_name=None, **kwargs):
+    def __init__(self, container, config=None, **kwargs):
         super(NewrelicPlugin, self).__init__()
+
         self.container = container
         self.container.error_hook.install(self.on_error)
         self.container.http_request_hook.install(self.on_http_request)
-        newrelic.agent.initialize(config_file, environment)
+        newrelic.agent.initialize(config.get('config_file'), config.get('environment'))
```
mouadino: config default is None, how can we do config.get(...) w/o ``if config`` ?

lymph/plugins/sentry.py
```diff
@@ -4,10 +4,10 @@
 
 
 class SentryPlugin(Plugin):
-    def __init__(self, container, dsn=None, **kwargs):
+    def __init__(self, container, config=None, **kwargs):
         super(SentryPlugin, self).__init__()
         self.container = container
-        self.client = Client(dsn)
+        self.client = Client(config.get('dsn'))
```
mouadino: ditto

lymph/plugins/newrelic.py
emulbreh: it can't really be None. It's just my old habbit to mark keyword arguments by providing a default.

docs/api/trace_api.rst
emulbreh: added a note

docs/api/trace_api.rst
emulbreh: I don't consider those two public api atm.

PR #295, opened 2016-01-25
Better error messages

mouadino:
Separate when discovery return empty service and when heartbeat fail to make
debugging easier.

### DIFF
--- a/lymph/core/rpc.py
+++ b/lymph/core/rpc.py
@@ -147,7 +147,9 @@ class ZmqRPCServer(Component):
     def _pick_instance(self, service):
         service.observe(services.REMOVED, self._on_service_instance_unavailable)
         choices = []
+        count = 0
         for instance in service:
+            count += 1
             try:
                 connection = self.connections[instance.endpoint]
             except KeyError:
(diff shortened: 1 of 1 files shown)


lymph/core/rpc.py
```diff
@@ -155,16 +157,18 @@ def _pick_instance(self, service):
                 continue
             if connection.is_alive():
                 choices.append(instance)
+        if count == 0:
+            raise NotConnected('servide have no instance')
         if not choices:
-            raise NotConnected('Not connected to %s' % service)
+            raise NotConnected('all %d instances connection are dead' % count)
```
emulbreh: Let's just include ``len(service)`` in the message

lymph/core/rpc.py
```diff
@@ -155,16 +157,18 @@ def _pick_instance(self, service):
                 continue
             if connection.is_alive():
                 choices.append(instance)
+        if count == 0:
+            raise NotConnected('servide have no instance')
```
promulo: 'service has' :)

lymph/core/rpc.py
promulo: well, since I'm catching typos: 'all %d instance connections' :D

mouadino: @emulbreh that should be the same as ``count`` ? it only that we don't have to loop over instances two times.

emulbreh: no need for this optimization, I'd rather save four lines.

mouadino: why ? it's the same in the end, just this solution is better optimized and w/o sacrificing readability ?

mei-li:
:+1:

PR #296, opened 2016-01-25
Log exception when register fail

mouadino:


lymph/discovery/zookeeper.py
```diff
@@ -140,7 +142,11 @@ def register(self, service_name, instance, timeout=1, namespace=SERVICE_NAMESPAC
             value.encode('utf-8'),
             ephemeral=True, makepath=True)
         # FIXME: result.set_exception(RegistrationFailure())
-        result.get(timeout=timeout)
+        try:
+            result.get(timeout=timeout)
+        except gevent.Timeout:
+            logger.exception('registeration of %s:%s failed', namespace, service_name)
```
emulbreh: registration

emulbreh:
We should probably reraise as a different exception and deal with it, but good enough for now (+1 except for the typo)

PR #298, opened 2016-01-27
Added nose2 plugin

emulbreh:


dushyant88:
I believe you also have to add nose2 to the requirements/dev.txt?

emulbreh:
only if we were to use nose2 by default now. I'd like to watch it for a while first, and also give pytest another chance.

dushyant88:
Cool, then +1

PR #299, opened 2016-01-29
Don't kill the entire CLI when some import fails

Drahflow:
Whenever something was wrong with any of the many services which registered CLI entry points, the entire lymph command line interface stopped working. This got annoying (and will get worse over time as more services are added).

This PR is a proposal for a fix of this situation.

### DIFF
--- a/lymph/cli/base.py
+++ b/lymph/cli/base.py
@@ -63,21 +63,6 @@ class Command(object):
         raise NotImplementedError
 
 
-class FailedToLoadCommand(Command):
-    def __init__(self, exc_type, exc_value, exc_traceback):
-        self.exc_type = exc_type
-        self.exc_value = exc_value
-        self.exc_traceback = exc_traceback
-        self.short_description = "Command unavailable: %s: %s" % (exc_type, exc_value)
-
-    @classmethod
-    def get_help(cls):
-        raise NotImplementedError("Command failed to load required module")
-
-    def run(self):
-        raise NotImplementedError("Command failed to load required module")
-
-
 _command_class_cache = None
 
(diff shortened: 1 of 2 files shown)


lymph/cli/base.py
```diff
@@ -74,9 +91,14 @@ def get_command_classes():
...
-            cls = entry_point.load()
-            cls.name = name
-            _command_class_cache[name] = cls
+            try:
+                cls = entry_point.load()
+                cls.name = name
+                _command_class_cache[name] = cls
+            except ImportError:
+                logger.error('Import error in %s, %s', entry_point, name)
+                traceback.print_exc()
```
emulbreh: ``print_exc()`` shouldn't be necessary here, ``error()`` already includes the traceback.
Let's include "command" and "entrypoint" in the error message to make clear why we tried to import.

lymph/cli/base.py
```diff
@@ -74,9 +91,14 @@ def get_command_classes():
...
+                _command_class_cache[name] = FailedToLoadCommand(sys.exc_type, sys.exc_value, sys.exc_traceback)
```
emulbreh: Let's just skip the command if we failed to load it. Less code, same information communicated.

Drahflow: I did this, because `lymph help` will just overflow the terminal space, scrolling the exception out of view. If the user searches for a specific command ("how was it just called?") the command would be gone without good explanation. This way it stays and the user can see that there is an error.

emulbreh:
Please remove FailedToLoadCommand, logging the exception is sufficient.

PR #300, opened 2016-02-05
Delete wait queue only when it exists

mouadino:
Seem like kombu delete "sometime" will raise a NotFound error when deleting
a queue that doesn't exists other time not, so we need to make sure that we
only delete queues on test method that create them.


emulbreh:
+1

PR #303, opened 2016-02-24
Set proctitle early

emulbreh:


stefano-dh:
:+1:  if travis

lymph/cli/service.py
```diff
@@ -77,6 +77,10 @@ def run(self):
 
         self._register_signals()
 
+        # set process title early since start() may take some time,
+        # even though endpoint may not be known yet
+        self._set_process_title()
+
         self.container.start(register=not self.args.get('--isolated', False))
 
         self._set_process_title()
```
eguven: forgot to remove?

emulbreh: proctitle currently contains the rpc endpoint, which is only known after start(). So we just call it again to replace the ``endpoint=None`` we got from the first call.

eguven: :+1:

PR #304, opened 2016-02-24
Without PR the _zmqpub logger breaks `lymph config`

Drahflow:


emulbreh:
The underlying problem here is that setup_logging mutates the configuration. This should be fixed instead by deep-copying the logging config.

Drahflow:
... however that has the effect (downside?) of not showing all log endpoints in `lymph config`. But I don't have a strong opinion on that.

emulbreh:
`lymph config` is meant to show what is read from config, not all defaults. Therefore I'd want to fix the logging setup and keep `config` output as-is.

PR #305, opened 2016-03-01
MAN-35: Add metric for service backend counts

Drahflow:


emulbreh:
This is almost definitely the wrong place to do this. If you just want a global instance count you should add it to lymph-monitor.

Drahflow:
It is a very good place to put it. Because it will report what the various instances *believe* to be the case -- and this is what needs to be correct (and hence monitored).

emulbreh:
let's rename it to something that communicates this idea then, `known_instances` maybe, with address as a tag.

